PR #5, opened 2023-03-18
Remove 'signMessage' project

bisgardo:
Replaced by https://github.com/Concordium/concordium-dapp-libraries/pull/29.


shjortConcordium (commented):
I don't think it makes sense to move this.

bisgardo:
> I don't think it makes sense to move this.

As mentioned in https://github.com/Concordium/concordium-dapp-libraries/pull/29#issuecomment-1475806068, I don't think this is a proper example dApp as it's only meant for testing.

PR #12, opened 2023-07-10
Add link to source code

DOBEN:
## Purpose

All hosted front ends should have a way to find the source code easily. Either through a link to an associated tutorial (that contains links to the source code) or if no tutorial exists through a link to the source code.

## Changes

- Add link to tutorial.
- Add `ChangeLog`.
- Add `version` to front end.
- Add testnet node link to command in ReadMe.md as an example.


gallery/src/Root.tsx
```diff
@@ -83,6 +84,19 @@ export default function Gallery() {
...
+            <div>
+                <br />
+                Version: {version} |{' '}
+                <a
+                    style={{ color: 'white' }}
+                    href="https://developer.concordium.software/en/mainnet/net/guides/gallery/index.html"
+                    target="_blank"
+                    rel="noreferrer"
+                >
+                    Explore the gallery tutorial here
```
abizjak, line 96: Shouldn't this end with a `.`?

abizjak (approved):
Not sure why the lock file had to change so much, but the link changes look good to me.

gallery/.yarnrc.yml
```diff
@@ -0,0 +1 @@
+nodeLinker: node-modules
```
shjortConcordium, line 1: I assume this is the cause of `yarn.lock` changes. :upside_down_face: 
(Why did you add this?)

DOBEN: It was automatically created by `yarn` (3.2.0 version) when I run `yarn install`. I think these changes in `yarn.lock` come from having a different `yarn` version.

DOBEN:
> Not sure why the lock file had to change so much, but the link changes look good to me.

It was automatically created by `yarn` (3.2.0 version) when I run `yarn install`. I think these changes in `yarn.lock` come from having a different `yarn` version.

gallery/.yarnrc.yml
shjortConcordium: Hmm, yes, I use yarn 1.22 (when it is not embedded in the repo like the sdk and browser wallet), so when i run `yarn` it changes the `yarn.lock` back.
I guess we need to specify which version is expected to be used. :thinking:

DOBEN: @shjortConcordium 

The reduced the `yarn` version in the `package.json` from `3.2.0` to `1.22.0`. This ensures that the lock files are not changes anymore.

https://github.com/Concordium/concordium-dapp-examples/blob/7f5e869c70799936622930204aba0332551325e4/gallery/package.json#L5

shjortConcordium (approved):

PR #17, opened 2023-07-19
Remove register public key tab

DOBEN:
## Purpose

addresses #https://github.com/Concordium/concordium-rust-smart-contracts/issues/258

The `public_key_registry` was removed from the sponsored transaction smart contract because after the protocol 6 update, we can query the public key from an account within the smart contract code. As a result, the front end does not need an extra tab anymore to register public keys for the user. This PR removes this register public key tab at the front end.

In addition, a new host function `check_account_signature` is available in the smart contract to verify signatures. The `check_account_signature` uses the `AccountSignatures` type as input parameter so this type is also used as input parameter for the `permit` function now.

## Changes

- Backend: 
     - Update to use the `AccountSignatures` type for the input parameter to the `permit` function.
     - Add TLS connection so we can use the internal stagenet node (which uses `https`). The protocol 6 is only live on stagenet at the moment.
- Frontend: 
     - Remove register public key tab.
     - Add esLinter.


sponsoredTransactions/backend/src/types.rs
```diff
@@ -129,3 +128,28 @@ pub struct Server {
...
+    #[concordium(size_length = 1)]
+    pub sigs: BTreeMap<CredentialIndex, CredentialSignatures>,
+}
+
+#[derive(Debug, Serial, Clone)]
+#[concordium(transparent)]
+pub struct CredentialSignatures {
+    #[concordium(size_length = 1)]
+    pub sigs: BTreeMap<KeyIndex, Signature>,
+}
```
abizjak, line 155: It's a bit unfortunate that we need to duplicate these definitions here.

If this is going to come up more often we should move all these definitions from concordium-std to contracts-common instead. 

The main reason they are in concordium-std is that the Signature type was already there.

I think we should move them instead of duplicating this here. An alternative would be to import the contract with its "PermitParams" as a dependency of the backend, but that has complication since they live in different repositories. 

Do you see any problems with moving htese definitions (and re-exporting them from concordium-std)? If not, could you do it?

sponsoredTransactions/frontend/src/SponsoredTransactions.tsx
```diff
@@ -146,25 +120,21 @@ async function calculateTransferMessage(nonce: string, tokenID: string, from: st
...
-                amount: '1',
-                data: '',
-                from: {
-                    Account: [from],
-                },
-                to: {
-                    Account: [to],
-                },
-                token_id: tokenID,
+    const transfer = [
```
abizjak, line 123: It's a bit annoying that you decided to reformat things in the same PR as making other changes. 

Could we try to avoid such things please?

sponsoredTransactions/frontend/src/SponsoredTransactions.tsx
```diff
@@ -241,23 +204,20 @@ async function calculateUpdateOperatorMessage(nonce: string, operator: string, a
...
-        toBuffer(SERIALIZATION_HELPER_SCHEMA, 'base64')
-    );
+    const serializedMessage = serializeTypeValue(message, toBuffer(SERIALIZATION_HELPER_SCHEMA, 'base64'));
 
     return serializedMessage;
 }
 
-async function getPublicKey(rpcClient: JsonRpcClient, account: string) {
+// eslint-disable-next-line  @typescript-eslint/no-explicit-any
+async function getPublicKey(rpcClient: any, account: string) {
```
abizjak, line 213: Why does this need to be "any'? Isn't there a tpye for the grpc client?

sponsoredTransactions/frontend/src/constants.ts
```diff
@@ -1,43 +1,34 @@
...
 export const SPONSORED_TX_CONTRACT_NAME = 'cis3_nft';
 
-export const SPONSORED_TX_CONTRACT_INDEX = 4376n;
+export const SPONSORED_TX_CONTRACT_INDEX = 5n;
 
-export const EXPIRY_TIME_SIGNATURE = '2030-08-08T05:15:00Z';
+export const CONTRACT_SUB_INDEX = 0n;
 
-export const UPDATE_OPERATOR_SCHEMA = 'EAEUAAIAAAAGAAAAdXBkYXRlFQIAAAAGAAAAUmVtb3ZlAgMAAABBZGQCCAAAAG9wZXJhdG9yFQIAAAAHAAAAQWNjb3VudAEBAAAACwgAAABDb250cmFjdAEBAAAADA';
+export const EXPIRY_TIME_SIGNATURE = '2030-08-08T05:15:00Z';
```
abizjak, line 17: Why is this hardcoded to an absolute time, even if it is one very far in the future? Why not use "now + 1h" or some relative time like that?

sponsoredTransactions/frontend/src/constants.ts
```diff
@@ -1,43 +1,34 @@
...
+export const VERIFIER_URL = '/api';
+
+export const REFRESH_INTERVAL = moment.duration(5, 'seconds');
 
 export const TESTNET_GENESIS_BLOCK_HASH = '4221332d34e1694168c2a0c0b3fd0f273809612cb13d000d5c2e00e85f50f796';
 
 export const SPONSORED_TX_CONTRACT_NAME = 'cis3_nft';
 
-export const SPONSORED_TX_CONTRACT_INDEX = 4376n;
+export const SPONSORED_TX_CONTRACT_INDEX = 5n;
```
abizjak, line 13: I guess this should be changed back for testnet deploy? Should this instead be a parameter?

sponsoredTransactions/backend/src/types.rs
DOBEN: Most of these types derive `Serialize` and `SchemaType` traits from `concordium-contracts-common-derive`. If I move them to `concordium-contracts-common` then it seems these derived traits are not available anymore since `concordium-contracts-common-derive` should not be used on its own but rather through `concordium-std`.

sponsoredTransactions/frontend/src/SponsoredTransactions.tsx
DOBEN: Totally agree. It was my mistake that I did some autoformatting in VS-Code and then later realized that the repo has no linting set up, so I either had to re-do all auto-formatting or set up linting. Sorry. I will watch out for it and try to avoid it in the future.

sponsoredTransactions/frontend/src/SponsoredTransactions.tsx
DOBEN: It is still the `JsonRPCClient` and I want to replace it in the next PR to GRPC. The `react-components` library is a newer version now and the  `JsonRPCClient` type exported from `web_sdk` was not compatible with it. I didn't bother to sort out the type since this part is deprecated and should be changed soon.

sponsoredTransactions/frontend/src/constants.ts
DOBEN: I updated it to be "now + 1 day". If ok, I choose 1 day because sometimes people travel and don't update their computer time immediately if only crossed a few timezones. For the demo, this should prevent people from getting errors at the front end due to slight time mismatches.  In production, I agree and would set it to something tighter e.g. 1h and give the user a warning if their time is off.

sponsoredTransactions/frontend/src/constants.ts
DOBEN: The smart contract index is now a build parameter.

sponsoredTransactions/frontend/src/constants.ts
DOBEN: https://github.com/Concordium/concordium-dapp-examples/issues/19

sponsoredTransactions/frontend/src/SponsoredTransactions.tsx
DOBEN: https://github.com/Concordium/concordium-dapp-examples/issues/19

sponsoredTransactions/frontend/src/constants.ts
DOBEN: We need to wait until protocol 6 is live on testnet:

- Remaining task in this issue https://github.com/Concordium/concordium-dapp-examples/issues/19:
Deploy sponsored transaction smart contract to testnet once protocol 6 is live. Update sponsored transaction smart contract index at the front end (package.json file). Update frontend/backend Read.me with the new index and related comments. Also update DockerFile.

sponsoredTransactions/backend/src/types.rs
abizjak: It is possible to work around that in an only slightly hacky way, like I did already in concordium_std where I do
```
use crate as concordium_std;
```

This makes `concordium_std` as a name available as the name for, in your case, ` contracts_common`. And since the exports of that crate are actually what you need it should work.

DOBEN: Types are now imported

abizjak (approved):

sponsoredTransactions/frontend/README.md
```diff
@@ -1,17 +1,18 @@
 # Sponsored Transactions Web Application
 
-The example project included in this repository serves as a working example of how to integrate sponsored transactions with smart contracts on the Concordium blockchain. This web app supports the following flows with the browser wallet (or wallet connect):
+The smart contracts, frontend, and backend works with protocol 6 which is live on stagenet. The SMART_CONTRACT_INDEX is set to 5 (stagenet) at the moment in the `package.json` file. It will be updated to testnet onece the protocol 6 is live on testnet.
```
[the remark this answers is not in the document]
DOBEN: removed that sentence altogether.

PR #34, opened 2023-12-20
Euroe demo

abizjak:
## Purpose

A demo example of a store using EUROe for payment.

## Checklist

- [ ] My code follows the style of this project.
- [ ] The code compiles without warnings.
- [ ] I have performed a self-review of the changes.
- [ ] I have documented my code, in particular the intent of the
      hard-to-understand areas.


euroe-demo/.prettierrc
```diff
@@ -0,0 +1,7 @@
+{
+  "trailingComma": "all",
+  "semi": true,
+  "singleQuote": true,
+  "printWidth": 120,
+  "bracketSpacing": true
+}
```
soerenbf, line 7: This is the usual config we use for formatting with prettier.

```suggestion
{
  "trailingComma": "all",
  "tabWidth": 4,
  "semi": true,
  "singleQuote": true,
  "printWidth": 120,
  "bracketSpacing": true
}
```

euroe-demo/.eslintrc.cjs
```diff
@@ -0,0 +1,11 @@
+module.exports = {
+  root: true,
+  env: { browser: true, es2020: true },
+  extends: ['eslint:recommended', 'plugin:@typescript-eslint/recommended', 'plugin:react-hooks/recommended'],
```
soerenbf, line 4: Would be good to:

```bash
yarn add -D eslint-config-prettier
```
and

```suggestion
  extends: ['eslint:recommended', 'plugin:@typescript-eslint/recommended', 'plugin:react-hooks/recommended', 'prettier'],
```

As this ensures that eslint and prettier don't conflict with regards to formatting rules.

euroe-demo/src/App.css
soerenbf, line 1: Can we just delete this?

soerenbf (approved):
Looks good, but consider the comments below.

euroe-demo/.prettierrc
abizjak: Hmm, I'm pretty sure I copied it from the dapp-template.

```
{
    "trailingComma": "all",
    "semi": true,
    "singleQuote": true,
    "printWidth": 120,
    "bracketSpacing": true
}
```

I'll add the tabwidth there to then.

euroe-demo/.eslintrc.cjs
abizjak: Ok. Added.

euroe-demo/src/App.css
abizjak: No, I like empty files :P

euroe-demo/.eslintrc.cjs
abizjak: Shouldn't this be added to the template as well?

soerenbf: It is actually already there ([here](https://github.com/Concordium/concordium-dapp-examples/blob/016a08f7943568349def7240e2104ec909be03ca/dapp-template/frontend/.eslintrc.cjs#L12C13-L12C13)), but it should state `'prettier'` instead of `'eslint-config-prettier'` according to the [documentation](https://github.com/Concordium/concordium-dapp-examples/blob/016a08f7943568349def7240e2104ec909be03ca/dapp-template/frontend/.eslintrc.cjs#L12C13-L12C13). I know this is my own doing, but if you could update it anyway, that would be nice 😄

euroe-demo/.prettierrc
soerenbf: Yes, sorry that'll be my mistake then 🙈

euroe-demo/.eslintrc.cjs
abizjak: Ok.

PR #38, opened 2024-01-19
Use metadata file with 0 decimals

### LINKED ISSUE #37: Sponsored Transactions Auction - Change metadata
**Description**
Currently the wCCD metadata is used for token metadata. Maybe reuse EUROe metadata or create new one.

In any case, change the number of decimals to be 0 so that it simplifies bidding in a way that we can bid only using int.

**Steps to Reproduce**
1.Mint 100 tokens.

**Expected Result**
100 tokens appear in the BW.

**Actual Result**
0.00100 tokens appear in the wallet.

**Versions**
- Software Version: 1.0.0

DOBEN:
## Purpose

closes #37 

## Changes

- Change metadata file to use a token with decimals 0. Token amounts in the wallets will be displayed as integers (no micro tokens such as 0.000001) which makes it easier for this demo app.
- Redeployed contracts to use the above metadata file for all token ids created.

### DIFF
--- a/sponsoredTransactionsAuction/frontend/src/constants.ts
+++ b/sponsoredTransactionsAuction/frontend/src/constants.ts
@@ -26,7 +26,8 @@ export const EPSILON_ENERGY = 200n;
 export const AUCTION_START = '2000-01-01T12:00:00Z'; // Hardcoded value for simplicity for this demo dApp.
 export const AUCTION_END = '2050-01-01T12:00:00Z'; // Hardcoded value for simplicity for this demo dApp.
 
-export const METADATA_URL = 'https://s3.eu-central-1.amazonaws.com/tokens.testnet.concordium.com/ft/wccd'; // We use the same metadat URL for every token_id for simplicity for this demo dApp. In production, you should consider using a different metadata file for each token_id.
+export const METADATA_URL =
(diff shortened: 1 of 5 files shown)


sponsoredTransactionsAuction/frontend/src/constants.ts
```diff
@@ -26,7 +26,8 @@ export const EPSILON_ENERGY = 200n;
 export const AUCTION_START = '2000-01-01T12:00:00Z'; // Hardcoded value for simplicity for this demo dApp.
 export const AUCTION_END = '2050-01-01T12:00:00Z'; // Hardcoded value for simplicity for this demo dApp.
 
-export const METADATA_URL = 'https://s3.eu-central-1.amazonaws.com/tokens.testnet.concordium.com/ft/wccd'; // We use the same metadat URL for every token_id for simplicity for this demo dApp. In production, you should consider using a different metadata file for each token_id.
+export const METADATA_URL =
+    'https://gist.githubusercontent.com/DOBEN/e035ef44705cdf8919f72c98a25d54eb/raw/8c6b375a2dff448e7bbd12a27fc420d41f268f12/gistfile1.txt'; // We use the same metadat URL for every token_id for simplicity for this demo dApp. In production, you should consider using a different metadata file for each token_id.
```
abizjak, line 30: It would be ideal if this was hosted somewhere in the Concordium organization, but it does not seem that's easily possible on github.

abizjak (approved):

sponsoredTransactionsAuction/frontend/src/constants.ts
DOBEN: let's keep this in mind and try to find something that is hosted by Concordium long-term. For now, I merge the PR.

PR #43, opened 2024-01-25
Hotfix node connection to use https instead of http

DOBEN:
## Purpose

Hotfix node connection to use https instead of http

## Changes

Node connection to use https instead of http


abizjak (approved):
But please make this a parameter in a followup.

PR #45, opened 2024-01-28
Add script to create and fill the track and trace contract with items

DOBEN:
## Purpose

A script to initialize a new track-and-trace contract and fill it with items. The status of the items will be updated by the script from `Produced` - > `InTransit` -> `InStore` -> `Sold`. This script helps to have the required contract with events that can be used to test the indexer (to be developed).

## Changes

Add script to create a track-and-trace contract and fill the contract with data for the indexer.


trackAndTrace/scripts/Cargo.toml
```diff
@@ -0,0 +1,13 @@
+[package]
+name = "track-and-trace-test"
+version = "0.1.0"
+edition = "2021"
+
+[dependencies]
+anyhow = "1.0"
+chrono = "0.4"
+clap = { version = "4.4", features = ["env", "derive"] }
+concordium-rust-sdk = { version = "4.0"}
```
abizjak, line 10: Please use the submodule.

trackAndTrace/scripts/README.md
```diff
@@ -0,0 +1,75 @@
...
+ 
+    You can generate the `track-and-trace.wasm.v1` file as follows:
+    - Navigate into the folder `https://github.com/Concordium/concordium-rust-smart-contracts/tree/main/examples/track-and-trace`.
+    - Compile the track-and-trace contract: `cargo concordium build -e -o ./concordium-out/track_and_trace.wasm.v1`.
+    - You can find the file `track_and_trace.wasm.v1` in `./concordium-out/track_and_trace.wasm.v1`.
+
+- `--admin-key-file` **directory** with one key in the
+  browser wallet export format. This key is used for all transactions executed by the script.
+
+- `--num-items` optional value to specify the number of items to be created in the track-and-trace smart contract. The default value is 1. The states of these items will be updated from 'Produced' -> 'InTransit' -> 'InStore' -> 'Sold'.
```
abizjak, line 20: I don't think we should be having defualts for arguments like this. This should always be a choice

trackAndTrace/scripts/README.md
```diff
@@ -0,0 +1,75 @@
...
+This will produce a single binary `track-and-trace-test` in `target/release` directory
+which can be run.
+
+
+## Run the script
+
+To run the script use for example the following command in this folder:
+
+```
+cargo run --release -- --node http://node.testnet.concordium.com:20000 --admin-key-file ./4SizPU2ipqQQza9Xa6fUkQBCDjyd1vTNUNDGbBeiRGpaJQc6qX.export --module ./track_and_trace.wasm.v1 --num-items 2
```
abizjak, line 45: I can't commend on the pictures, but I am not in favor of duplicating "how to export keys from wallet" here.

Screenshots get obsolete quickly. Please just link to the developer documentation. If that data is not there yet add it there instead.

trackAndTrace/scripts/src/main.rs
```diff
@@ -0,0 +1,271 @@
...
+use std::sync::Arc;
+mod types;
+use types::*;
+
+/// Command line configuration of the application.
+#[derive(Debug, clap::Parser)]
+struct Args {
+    #[arg(
+        long = "node",
+        default_value = "http://node.testnet.concordium.com:20000",
```
abizjak, line 24: Please use `https://grpc.testnet.concordium.com` instead since `node` is intended to be deprecated.

trackAndTrace/scripts/src/main.rs
```diff
@@ -0,0 +1,271 @@
...
+
+    let mut client = sdk::Client::new(endpoint)
+        .await
+        .context("Unable to establish connection to the node.")?;
+
+    // Load account keys and sender address from a file
+    let keys: WalletAccount = WalletAccount::from_json_file(args.admin_keys_path)
+        .context("Could not read the keys file.")?;
+
+    let admin_key = Arc::new(keys);
```
abizjak, line 72: Why `Arc`? Does not seem needed.

trackAndTrace/scripts/src/types.rs
```diff
@@ -0,0 +1,48 @@
...
+pub enum Status {
+    /// Item is produced.
+    Produced,
+    /// Item is in transit.
+    InTransit,
+    /// Item is in store.
+    InStore,
+    /// Item is sold.
+    Sold,
+}
```
abizjak, line 27: Why are these types not defined in the contract instead?

trackAndTrace/scripts/src/types.rs
```diff
@@ -0,0 +1,48 @@
+use concordium_rust_sdk::smart_contracts::{common as concordium_std, common::Serial};
+
+/// The location of the metadata URL and an optional hash of the content.
+#[derive(Serial)]
+pub struct MetadataUrl {
+    /// The URL following the specification RFC1738.
+    pub url:  String,
+    /// An optional hash of the content.
+    pub hash: Option<[u8; 32]>,
+}
```
abizjak, line 10: Why is this not imported from concordium-std or from the contract?

trackAndTrace/scripts/src/types.rs
```diff
@@ -0,0 +1,48 @@
...
+#[derive(Serial)]
+pub struct ChangeItemStatusParamsByAdmin {
+    /// The item's id.
+    pub item_id:         ItemID,
+    /// The item's new status.
+    pub new_status:      Status,
+    /// Any additional data encoded as generic bytes. Usecase-specific data can
+    /// be included here such as temperature, longitude, latitude, ... .
+    pub additional_data: AdditionalData,
+}
```
abizjak, line 46: Isn't this meant to be defined in the contract? Why is it duplicated here?

trackAndTrace/scripts/src/types.rs
```diff
@@ -0,0 +1,48 @@
...
+    Sold,
+}
+
+/// Partial type of the type `ChangeItemStatusParamsByAdmin`.
+#[derive(Serial)]
+pub struct AdditionalData {
+    /// Any additional data encoded as generic bytes. Usecase-specific data can
+    /// be included here such as temperature, longitude, latitude, ... .
+    pub bytes: Vec<u8>,
+}
```
abizjak, line 35: This is defined in our libraries already.

abizjak (changes requested):

trackAndTrace/scripts/src/main.rs
abizjak: So the `grpc.testnet.concordium.com` and `grpc.mainnet.concordium.software` are the public grpc endpoints of the node. I don't know why the names were decided anymore, but those are the links. Let's not revisit the naming again.

There are technical complications with https on the `node` URL. We are aware of annoyances with deprecating it so we're not just going to shut it off, but it will be removed eventually. So please use the link I suggested.

Bargsteen (approved):
LGTM

trackAndTrace/scripts/src/main.rs
```diff
@@ -0,0 +1,269 @@
...
+            anyhow::bail!("Update item status failed: {err:#?}");
+        }
+    }
+
+    // Update items from `InStore` to `Sold`
+    for i in 0..args.num_items {
+        let param: ChangeItemStatusParamsByAdmin = ChangeItemStatusParamsByAdmin {
+            item_id:         i as u64,
+            new_status:      Status::Sold,
+            additional_data: AdditionalData { bytes: vec![] },
```
abizjak, line 243: Perhaps it's worth adding a `Default` implementation for AdditoinalData to make this a tiny bit nicer.

trackAndTrace/scripts/README.md
```diff
@@ -0,0 +1,71 @@
+## Test script to facilitate deploying and filling the track and trace contract with data for the indexer.
+
+This package contains a test script to facilitate filling the track and trace contract with data for the indexer.
```
abizjak, line 3: It's a bit confusing to me to have this script.

I would expect a script to create a new instance of track and trace with perhaps some new items.

But doing transitions in the script seems odd to me for this reference project.

If something like this is needed for a test I would recommend putting it as an integration test in the smart contract or in some other folder explicitly marked as a test folder.

And these scripts would be just deployment and management scripts that are a normal part of the reference project.

trackAndTrace/scripts/src/main.rs
```diff
@@ -0,0 +1,269 @@
...
+            "Submitted update item status with index {i} to `InTransit` in transaction hash \
+             {tx_hash}"
+        );
+
+        if let Err(err) = tx_hash.wait_for_finalization().await {
+            anyhow::bail!("Update item status failed: {err:#?}");
+        }
+    }
+
+    // Update items from `InTransit` to `InStore`
```
abizjak, line 209: I think the script generally looks good, but I don't agree with its scope, namely having this part with transitions.

I believe that should be a separate test.

trackAndTrace/scripts/README.md
DOBEN: Moved it into a `test_scripts` folder.

trackAndTrace/scripts/src/main.rs
DOBEN: The script facilitates to development (as written in the Read.me) of the indexer and as such it is useful to fill the contract with data. 

If in addition, if we want to have an extra deployment script then I agree we should add that to the `smart-contract` folder (hasn't been written yet).

abizjak: So I don't really think we should have this and a separate deployment script that will do most of the same things.

We'll end up with two scripts to maintain, where 80% of the code is almost literal copy paste between them.

I also don't understand the rationale for design.

Assuming you had the deployment script already, then for testing the indexer I would expect you would want something that makes a few transitions over time. So it'd just look up some items and make those transitions and keep doing them.

That would make more sense to me.

trackAndTrace/test-scripts/src/main.rs
```diff
@@ -0,0 +1,287 @@
...
+        short = 'i',
+        help = "A JSON file containing the input parameter."
+    )]
+    input_parameter_json_file: std::path::PathBuf,
+    #[arg(
+        long = "num-items",
+        short = 'i',
+        help = "Number of items to be created in the contract."
+    )]
+    num_items:                 usize,
```
abizjak, line 48: It seems like if you set this to be a u64 instead you can avoid `i as u64` conversions.

trackAndTrace/test-scripts/src/main.rs
```diff
@@ -0,0 +1,287 @@
...
+            eprintln!("Send deployment transaction with hash {hash}.");
+
+            let (block_hash, result) = client.wait_until_finalized(&hash).await?;
+
+            if let Some(err) = result.is_rejected_account_transaction() {
+                anyhow::bail!("Failed to deploy module: {err:#?}");
+            }
+
+            eprintln!("Deployed module with reference {mod_ref} in block {block_hash}.");
+        }
```
abizjak, line 120: You could simplify this with the latest rust SDK by using the ModuleDeployBuilder
```rust
            let module =
                WasmModule::from_file(&module_path).context("Could not read contract module.")?;
            let builder =
                ModuleDeployBuilder::dry_run_module_deploy(client, keys.address, module).await?;
            let handle = builder.send(&keys).await?;
            println!("Module deployment transaction {handle} submitted.");
            let result = handle.wait_for_finalization().await?;
            println!("Module {} deployed.", result.module_reference);
```

trackAndTrace/test-scripts/src/main.rs
```diff
@@ -0,0 +1,287 @@
...
+            anyhow::bail!("Failed to initialize contract: {error:#?}.");
+        }
+        let info = result.contract_init().context("Expect an init result")?;
+
+        eprintln!(
+            "Successfully initialized contract in block {block_hash}, with address {}.",
+            info.address
+        );
+
+        contract_client::ContractClient::<TrackAndTraceContract>::create(
```
abizjak, line 169: You could simplify this by using the ContractInitBuilder
```rust
            let builder = ContractInitBuilder::<WeatherContractMarker>::dry_run_new_instance(
                client,
                account.address,
                mod_ref,
                "track_and_trace",
                Amount::zero(),
                &params,
            )
            .await?;
            println!(
                "The maximum amount of NRG allowed for the transaction is {}.",
                builder.current_energy()
            );
            let handle = builder.send(&account).await?;
            println!("Transaction {handle} submitted. Waiting for finalization.");
            let (contract_client, events) = handle.wait_for_finalization().await?;
            println!(
                "Initialized a new smart contract instance at address {}.",
                contract_client.address
            );

abizjak (approved):

PR #46, opened 2024-01-29
Add track and trace smart contract

DOBEN:
## Purpose

The track-and-trace contract is modeled based on a state machine. The flow of the state machine is as follows: The `ADMIN` creates a new item with status `Produced`. Each new item is assigned the `next_item_id`. The `next_item_id`
value is sequentially increased by 1 in the contract's state. In the test cases, the different
roles (`PRODUCER`, `TRANSPORTER`, `SELLER`) can update the item's status based on the rules of the state machine
as follows: `Produced -> InTransit -> InStore -> Sold`. The `ADMIN` role is an exception and can set an item's status to any value at any time ignoring the state machine rules.

Moved from https://github.com/Concordium/concordium-rust-smart-contracts/pull/388

## Changes

Add track and trace smart contract.


trackAndTrace/README.md
```diff
@@ -0,0 +1 @@
+## Track and Trace Project
```
abizjak, line 1: Is this meant to be empty?

trackAndTrace/smart-contract/Cargo.toml
```diff
@@ -0,0 +1,27 @@
...
+default = ["std", "wee_alloc"]
+std = ["concordium-std/std", "concordium-cis2/std"]
+wee_alloc = ["concordium-std/wee_alloc"]
+
+[dependencies]
+concordium-std = { version = "9.0", default-features = false }
+concordium-cis2 = { version = "6.0", default-features = false }
+
+[dev-dependencies]
+concordium-smart-contract-testing = { version = "4.0" }
```
abizjak, line 19:
```suggestion
concordium-smart-contract-testing = { version = "4.1" }
```

trackAndTrace/smart-contract/src/lib.rs
```diff
@@ -0,0 +1,698 @@
...
+/// The ItemCreatedEvent is logged when an item is created.
+#[derive(Serialize, SchemaType, Debug, PartialEq, Eq)]
+pub struct ItemCreatedEvent {
+    /// The item's id.
+    pub item_id:      ItemID,
+    /// The item's metadata_url.
+    pub metadata_url: Option<MetadataUrl>,
+}
+
+/// The ItemStatusChangedEvent is logged when the status of an item is updated.
```
abizjak, line 56:
```suggestion
/// The [`ItemStatusChangedEvent`] is logged when the status of an item is updated.
```

trackAndTrace/smart-contract/src/lib.rs
```diff
@@ -0,0 +1,698 @@
...
+#[derive(Serialize, SchemaType, Debug, PartialEq, Eq)]
+pub struct ItemStatusChangedEvent {
+    /// The item's id.
+    pub item_id:         ItemID,
+    /// The item's new status.
+    pub new_status:      Status,
```
abizjak, line 62: Would it make sense to log the old and the new status? For easier tracking?

trackAndTrace/smart-contract/src/lib.rs
```diff
@@ -0,0 +1,698 @@
...
+#[derive(Serialize, PartialEq, Eq, Reject, SchemaType, Clone, Copy, Debug)]
+pub enum Roles {
+    /// Admin role.
+    ADMIN,
+    /// Producer role.
+    PRODUCER,
+    /// Transporter role.
+    TRANSPORTER,
+    /// Seller role.
+    SELLER,
```
abizjak, line 105: Can you please use the normal convention for enums, and use CamelCase naming.

trackAndTrace/smart-contract/src/lib.rs
```diff
@@ -0,0 +1,698 @@
...
+    InTransit,
+    /// Item is in store.
+    InStore,
+    /// Item is sold.
+    Sold,
+}
+
+impl Status {
+    fn new() -> Self { Status::Produced }
+}
```
abizjak, line 123: I think it's odd to have this method, especially since the behaviour is not obvious. Why is "new" produced?

If you want a method then name it `proceed` to make it explicit.

trackAndTrace/smart-contract/src/lib.rs
```diff
@@ -0,0 +1,698 @@
...
+
+// Implementing a custom schemaType for the `Event` struct containing all
+// events. This custom implementation flattens the fields to avoid one
+// level of nesting. Deriving the schemaType would result in e.g.: {"Nonce":
+// [{...fields}] }. In contrast, this custom schemaType implementation results
+// in e.g.: {"Nonce": {...fields} }
+impl schema::SchemaType for Event {
```
abizjak, line 130: For this reference project I think this is just needless complexity.

You get a bit more nesting with the generated code, but the benefit is that the code looks much more accessible.

I don't see the added nesting as a bit issue to anybody.

trackAndTrace/smart-contract/src/lib.rs
```diff
@@ -0,0 +1,698 @@
...
+    /// Failed logging because the log is full.
+    LogFull, // -2
+    /// Failed logging because the log is malformed.
+    LogMalformed, // -3
+    /// Failed because the actor is not authorized to invoke the entry point.
+    Unauthorized, // -4
+    /// Failed to revoke role because it was not granted in the first place.
+    RoleWasNotGranted, // -5
+    /// Failed to grant role because it was granted already in the first place.
+    RoleWasAlreadyGranted, // -6
```
abizjak, line 220: I'm wondering what the rationale is for having this as an error?

trackAndTrace/smart-contract/src/lib.rs
```diff
@@ -0,0 +1,698 @@
...
+#[derive(Serialize, SchemaType, PartialEq, Eq, Debug)]
+pub struct ViewState {
+    /// The next item id that will be assigned to an item when the admin creates
+    /// it.
+    pub next_item_id: ItemID,
+    /// A vector containing all roles granted to addresses.
+    pub roles:        Vec<(Address, Vec<Roles>)>,
+    /// A vector containing all items with their states.
+    pub items:        Vec<(ItemID, ItemState)>,
+}
```
abizjak, line 327: I see that this is convenient for debugging, and we have this kind of thing in other contracts.

But this is not really a sustainable thing, because the list of items can grow arbitrarily large. So this method will eventually stop working.

What we should have is to view the state of an item or a list of items that you supply.

I would also expect to have a view method for the list of roles, but that would be separate from viewing items.

trackAndTrace/smart-contract/src/lib.rs
```diff
@@ -0,0 +1,698 @@
...
+pub struct AdditionalData {
+    /// Any additional data encoded as generic bytes. Usecase-specific data can
+    /// be included here such as temperature, longitude, latitude, ... .
+    pub bytes: Vec<u8>,
+}
+
+/// The parameter type for the contract function `changeItemStatusByAdmin` which
+/// updates the status of an item.
+#[derive(Serialize, SchemaType)]
+pub struct ChangeItemStatusParamsByAdmin {
```
abizjak, line 427: This type seems to be the same as `ItemStatusChangedEvent`. Why do we have two? It adds cost to have this.

trackAndTrace/smart-contract/src/lib.rs
```diff
@@ -0,0 +1,698 @@
...
+/// - It fails to log the `ItemStatusChangedEvent`.
+#[receive(
+    contract = "track_and_trace",
+    name = "changeItemStatusByAdmin",
+    parameter = "ChangeItemStatusParamsByAdmin",
+    error = "CustomContractError",
+    mutable,
+    enable_logger
+)]
+fn change_item_status_by_admin(
```
abizjak, line 454: So I'm not sure why we have this method.

I thought we're gonna have a generic state machine setup that is initialized when the contract is created.

That setup **could** be that the admin (a specific role) can always do arbitrary state transitions. But it does not have to be.

DOBEN: From our discussion with the product team, the two requirements they wanted to include:
- The admin can change the state of an item freely to any value.
- The other roles can change the item's state based on a `state-machine` rule design (usually a linear supply chain).

Having a general state machine that can be arbitrarily initiated when initializing a new instance of the contract seems to result in rather complex code and quite expensive logic (a basic linear supply chain is the most common case in such track-and-trace applications, not arbitrary state machine transitions).

Our initial approach to having just one entrypoint, for admin and the other roles, to change the state of the item, results in 
that the redundant input parameters `new_state` needs to be tracked. For all the other roles it is clear what the `next_state` is given a state machine (so no need to add it as an input_paramer). To adhere to smart contract development best practices: Keep the smart contract state small and keep input parameters, and events small to avoid blockchain bloat, there are now two entrypoints to have a very efficient/cheap way to update the state of an item by the other roles.

trackAndTrace/smart-contract/src/lib.rs
DOBEN: Since the `old_state` is always the previous `new_state` from the previous event associated with the item, it seems redundant to track it again. The front end displays a new data point (with the `new_state` when an event is logged).

https://concordium.atlassian.net/wiki/spaces/EN/pages/1289224193/SCM+on+Concordium+Reference+Project+V0.4?atlOrigin=eyJpIjoiNTc4ZTBhMjAtODkzNC0xMzQxLTRmZmEtZGMwMzI5MjA0Zjg4IiwicCI6ImMifQ&atl_p=eyJpIjoiNTc4ZTBhMjAtODkzNC0xMzQxLTRmZmEtZGMwMzI5MjA0Zjg4IiwidCI6InRvcExlZnRWaWV3Q29udGFpbmVyIiwic291cmNlIjoiZW1haWwifQ

trackAndTrace/smart-contract/src/lib.rs
DOBEN: removed

trackAndTrace/smart-contract/src/lib.rs
DOBEN: removed

trackAndTrace/smart-contract/src/lib.rs
DOBEN: It does not make sense to grant a role that is already granted (waste of money that has to be paid as transaction fee). If we revert in that case, then it can be caught via dry running the transaction.

trackAndTrace/smart-contract/src/lib.rs
DOBEN: Using a type alias now:
```
/// The parameter type for the contract function `changeItemStatusByAdmin` which
/// updates the status of an item. The type is equivalent to the `ItemStatusChangedEvent` type.
type ChangeItemStatusParamsByAdmin = ItemStatusChangedEvent;
```

DOBEN: Added view functions.

trackAndTrace/smart-contract/src/lib.rs
abizjak: Yes, it should not be necessary to do so. But I thought it would be simpler when tracking to have this.

trackAndTrace/smart-contract/src/lib.rs
abizjak: I think this is specializing the contract way too much. The linear progression also seems quite a special case, and I expect that there can be choice in many states, and it is the owner of the item in a given state that should decide where to transition the item.

I don't really understand the argument about complexity. What I had in mind is a state machine like this
```rust
/// Your smart contract state.
#[derive(Serial, DeserialWithState)]
#[concordium(state_parameter = "S")]
pub struct Transitions<State: Serialize, S> {
    map: StateMap<State, StateMap<AccountAddress, StateSet<State, S>, S>, S>,
}

impl<State: Serialize, S: HasStateApi> Transitions<State, S> {
    /// Add a new transition. Return if the transition was fresh.
    pub fn add(
        &mut self,
        builder: &mut StateBuilder<S>,
        from: State,
        address: AccountAddress,
        to: State,
    ) -> bool {
        let mut transition = self.map.entry(from).or_insert_with(|| builder.new_map());
        let mut targets = transition
            .entry(address)
            .or_insert_with(|| builder.new_set());
        targets.insert(to)
    }

    /// Remove a transition and return if it was present.
    pub fn remove(&mut self, from: &State, address: &AccountAddress, to: &State) -> bool {
        let Some(mut transition) = self.map.get_mut(from) else {
            return false;
        };
        let Some(mut targets) = (&mut *transition).get_mut(address) else {
            return false
        };
        (&mut *targets).remove(to)
    }

    /// Check if a transition is allowed.
    pub fn check(&self, from: &State, address: &AccountAddress, to: &State) -> bool {
        let Some(transition) = self.map.get(from) else {
            return false;
        };
        let Some(targets) = transition.get(address) else {
            return false
        };
        targets.contains(to)
    }

    pub fn from_iter(
        state_builder: &mut StateBuilder<S>,
        i: impl IntoIterator<Item = (State, AccountAddress, State)>,
    ) -> Self {
        let mut r = Self {
            map: state_builder.new_map(),
        };
        for (from, address, to) in i {
            r.add(state_builder, from, address, to);
        }
        r
    }
}
```
Which seems extremely simple to me, and simpler than hardcoding the admin as a special role.

As to complexity wise, this will indeed mean that initialization of a new contract can be a bit more expensive, depending on how many transitions we have, but checking if a transition is allowed will not be any more expensive computation wise.

DOBEN: updated the design of the state machine.

trackAndTrace/smart-contract/src/lib.rs
DOBEN: Type got removed after changing the state machine design.

trackAndTrace/smart-contract/src/lib.rs
```diff
@@ -0,0 +1,645 @@
...
+    ensure!(
+        state.has_role(&sender, Roles::Admin),
+        CustomContractError::Unauthorized
+    );
+
+    // Check that the `address` had previously not held the specified role.
+    ensure!(
+        !state.has_role(&params.address, params.role),
+        CustomContractError::RoleWasAlreadyGranted
+    );
```
abizjak, line 577: Why is it important to have this check?

abizjak (approved):

trackAndTrace/smart-contract/src/lib.rs
```diff
@@ -0,0 +1,645 @@
...
+    };
+
+    // Update the state of the item.
+    let old_item_status = item.status;
+    item.status = param.new_status;
+    drop(item);
+
+    let verify = host
+        .state()
+        .check(&old_item_status, &account, &param.new_status);
```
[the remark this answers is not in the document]
DOBEN: Yup, I was struggling here with mutable references quite a bit. 

If there are some complex constructs then yes it is rather `weird` in Concordium contracts, that even if we want to adhere to smart contract best practice (first do all the checks, then do state changes), we end up with the opposite order due to mutable references. I don't have a solution for it but I agree that the order looks rather weird at first glance and people not familiar with Rust.

trackAndTrace/smart-contract/src/lib.rs
DOBEN: I think it might be beneficial to have this revert so we can dry-run the transaction, but not feeling strongly about it. If we remove it the smart contract module is a bit smaller which saves gas, so I now removed it.

PR #47, opened 2024-02-01
Indexer

DOBEN:
## Purpose

addresses https://github.com/Concordium/concordium-dapp-examples/issues/50

Next: A follow-up PR should include the http server that provides two entrypoints for the front end to query the database.

## Changes

- Add event indexer for the `track-and-trace` project.


trackAndTrace/indexer/src/main.rs
```diff
@@ -0,0 +1,476 @@
...
+                            connection from pool",
+                                    ) {
+                                        Ok(db) => db,
+                                        Err(e) => {
+                                            receiver.close();
+                                            return Err(e);
+                                        }
+                                    };
+                                    true
+                                }
```
abizjak, line 378: I really don't like this duplication very much. It would be better to organize this like it is in the bridge indexer

trackAndTrace/indexer/src/main.rs
```diff
@@ -0,0 +1,476 @@
...
+    spinner.finish_and_clear();
+    shutdown_handle.abort();
+
+    Ok(())
+}
+
+/// Inserts data related to the `ItemCreatedEvent` into the database.
+/// Everything is commited as a single transactions allowing
+/// for easy restoration from the last recorded checkpoint into
+/// the database. Returns the duration it took to process the transaction.
```
abizjak, line 397: The problem is that this is not really true. The easy restoration would be much clearer and easier if we inserted in one DB transaction per block.

abizjak (changes requested):

trackAndTrace/indexer/src/main.rs
```diff
@@ -0,0 +1,476 @@
...
+
+    // Set up progress bar.
+    let spinner =
+        ProgressBar::new_spinner().with_style(ProgressStyle::with_template("{spinner} {msg}")?);
+
+    // Database process runs until the stop flag is triggered.
+    let stop_flag = Arc::new(AtomicBool::new(false));
+    let shutdown_handle = tokio::spawn(set_shutdown(stop_flag.clone()));
+
+    let mut contrat_set = BTreeSet::new();
```
soerenbf, line 212:
```suggestion
    let mut contract_set = BTreeSet::new();
```

trackAndTrace/indexer/src/main.rs
```diff
@@ -0,0 +1,476 @@
...
+            .tls_config(tonic::transport::channel::ClientTlsConfig::new())
+            .context("Unable to construct TLS configuration for the Concordium API.")?
+    } else {
+        app.node_endpoint
+    }
+    .connect_timeout(std::time::Duration::from_secs(5))
+    .timeout(std::time::Duration::from_secs(10));
+
+    // Establish connection to the postgres database.
+    let db_pool = DatabasePool::create(app.db_connection.clone(), 2, true)
```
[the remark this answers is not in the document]
soerenbf: Maybe I can provide some clarity as to why `2`. This is what I put for the indexer for the GC election project as well. My train of thought was that when attempting to reconnect to the DB you'd have 2 connections until the first one would be dropped (maybe this assumption is wrong, but I thought it couldn't hurt to allow 2 connections for this purpose).

trackAndTrace/indexer/resources/schema.sql
```diff
@@ -0,0 +1,41 @@
+-- Table containing server settings
+CREATE TABLE IF NOT EXISTS settings (
+  id BOOL PRIMARY KEY DEFAULT true CHECK (id), -- To constrain table to have a single row.
+  latest_block_height INT8,
+  latest_transaction_hash BYTEA,
+  latest_event_index INT8,
```
soerenbf, line 6: This would also be redundant if we commit per block

trackAndTrace/indexer/resources/schema.sql
```diff
@@ -0,0 +1,41 @@
...
+  contract_index INT8 NOT NULL,
+  contract_subindex INT8 NOT NULL
+);
+
+-- Table containing item_status_changed_events successfully submitted to the database from the contract monitored.
+CREATE TABLE IF NOT EXISTS item_status_changed_events (
+  id INT8 PRIMARY KEY,
+  block_height INT8 NOT NULL,
```
soerenbf, line 14: Why is block height interesting? If it is to make it possible to order the data by something from the chain, would it not be better to add the block time instead, as it would probably be more interesting for the consumer of the data?

trackAndTrace/indexer/resources/schema.sql
```diff
@@ -0,0 +1,41 @@
...
+  new_status INT8 NOT NULL,
+  additional_data BYTEA NOT NULL,
+  -- Define composite unique constraint to ensure each event exists at most once in the database
+  CONSTRAINT unique_item_status_changed UNIQUE (block_height, transaction_hash, event_index)
+);
+
+-- Table containing item_created_events successfully submitted to the database from the contract monitored.
+CREATE TABLE IF NOT EXISTS item_created_events (
+  id INT8 PRIMARY KEY,
+  block_height INT8 NOT NULL,
```
soerenbf, line 27: Also here

trackAndTrace/indexer/resources/schema.sql
```diff
@@ -0,0 +1,41 @@
...
+  item_id INT8 NOT NULL,
+  new_status INT8 NOT NULL,
+  additional_data BYTEA NOT NULL,
+  -- Define composite unique constraint to ensure each event exists at most once in the database
+  CONSTRAINT unique_item_status_changed UNIQUE (block_height, transaction_hash, event_index)
+);
+
+-- Table containing item_created_events successfully submitted to the database from the contract monitored.
+CREATE TABLE IF NOT EXISTS item_created_events (
+  id INT8 PRIMARY KEY,
```
limemloh, line 26: **Suggestion:** It seems that we are incrementing this ID manually in the insert query, consider using [BIGSERIAL](https://www.postgresql.org/docs/current/datatype-numeric.html) instead

trackAndTrace/indexer/src/main.rs
```diff
@@ -0,0 +1,476 @@
...
+        sender,
+    ));
+
+    // The indexer starts processing historical events and then listens for new
+    // events that are coming in as the blockchain progresses.
+    while let Some((block, contract_update_infos)) = receiver.recv().await {
+        // Stop indexer when triggered.
+        if stop_flag.load(Ordering::Acquire) {
+            break;
+        }
```
limemloh, line 232: **Suggestion:** Signalling the service to stop, will not actually stop until the indexer receives a new block, which is not the end of the world, but not ideal.

```suggestion
    let shutdown_signal = set_shutdown();
    // The indexer starts processing historical events and then listens for new
    // events that are coming in as the blockchain progresses.
    loop {
        let (block, contract_update_infos) = tokio::select! {
            biased;
            _ = shutdown_signal => { break },
            Some(v) = receiver.recv() => v,
            else => { break }
        };
```
The above is not really tested, but it should remove the need for `stop_flag` as well

trackAndTrace/indexer/src/main.rs
```diff
@@ -0,0 +1,476 @@
...
+                                    db = match db_pool.get().await.context(
+                                        "Failed to get new database
+                            connection from pool",
+                                    ) {
+                                        Ok(db) => db,
+                                        Err(e) => {
+                                            receiver.close();
+                                            return Err(e);
+                                        }
+                                    };
```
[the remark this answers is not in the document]
abizjak: It's just that there is no point with having the pool for the indexer. So I don't like having things that don't have a point.

trackAndTrace/indexer/resources/schema.sql
```diff
@@ -0,0 +1,41 @@
...
+  latest_block_height INT8,
+  latest_transaction_hash BYTEA,
+  latest_event_index INT8,
+  contract_index INT8 NOT NULL,
+  contract_subindex INT8 NOT NULL
+);
+
+-- Table containing item_status_changed_events successfully submitted to the database from the contract monitored.
+CREATE TABLE IF NOT EXISTS item_status_changed_events (
+  id INT8 PRIMARY KEY,
```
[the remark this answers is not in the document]
abizjak: It would be better to do it explicitly since it will probably be leaked in queries for pagination, etc.

SERIAL does not guarantee that there are no gaps, and you can see in practice if transactions are rolled back for example that there will be gaps in IDs.

While this does not matter for storage, these IDs will likely be exposed for pagination purposes in queries, so some clients are bound to rely on them. Since it's easy to ensure they consistent and deterministic, we should do so.

trackAndTrace/indexer/resources/schema.sql
```diff
@@ -0,0 +1,41 @@
...
+  block_height INT8 NOT NULL,
+  transaction_hash BYTEA NOT NULL,
+  event_index INT8 NOT NULL,
+  item_id INT8 NOT NULL,
+  new_status INT8 NOT NULL,
+  additional_data BYTEA NOT NULL,
+  -- Define composite unique constraint to ensure each event exists at most once in the database
+  CONSTRAINT unique_item_status_changed UNIQUE (block_height, transaction_hash, event_index)
```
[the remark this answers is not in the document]
DOBEN: It is not unique if we want to refer to a specific event in a general smart contract. Since the indexer should be generic enough to apply to any contract, this is tracked here.

For our specific smart contract example: yes the transaction hash would be enough to uniquely identify an event.

trackAndTrace/indexer/src/main.rs
```diff
@@ -0,0 +1,476 @@
...
+        global = true,
+        env = "CCD_INDEXER_NODE"
+    )]
+    node_endpoint:    concordium_rust_sdk::v2::Endpoint,
+    #[arg(
+        long = "start",
+        short = 's',
+        help = "The start time when the track and trace contract was initialized. The format is \
+                ISO-8601, e.g. 2024-01-23T12:13:14Z.",
+        env = "CCD_INDEXER_START"
```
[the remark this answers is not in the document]
DOBEN: It is the block height now.

trackAndTrace/indexer/resources/schema.sql
DOBEN: I added the `block_time` to the database, which is now returned when querying the rows of the database. In addition, I kept the `block_height` in the database because it will allow to figure out which is the last `block_height` that was processed to be able to re-start the indexer.

The indexer starts by providing it the `block_height` to start from. The `block_time` is hard to copy-paste from CCDScan in the way it is presented at the moment.

trackAndTrace/indexer/src/main.rs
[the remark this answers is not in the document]
DOBEN: I got inspiration from @soerenbf's great work on the GC election project :). 

Nonetheless, I changed the value to 1 now since we commit one DB transaction per block now without any re-try logic.

trackAndTrace/indexer/src/main.rs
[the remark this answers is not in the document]
DOBEN: As Soeren already mentioned, the `http` server should access the DB as well through the pool. Next time, it should be added in the next PR, agree. For now, I kept the DB pool in this PR so.

trackAndTrace/smart-contract/Cargo.toml
```diff
@@ -8,12 +8,14 @@ description = "A smart contract for tracking and tracing items on a supply chain
...
-std = ["concordium-std/std", "concordium-cis2/std"]
+std = ["concordium-std/std"]
 wee_alloc = ["concordium-std/wee_alloc"]
+derive-serde = ["concordium-contracts-common/derive-serde", "serde"]
 
 [dependencies]
 concordium-std = { version = "9.0", default-features = false }
-concordium-cis2 = { version = "6.0", default-features = false }
+serde = { version = "1", optional = true, features = ["derive"] }
+concordium-contracts-common = "9.0.0"
```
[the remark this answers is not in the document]
DOBEN: It is needed because the feature `derive-serde` includes `concordium-contracts-common/derive-serde`. `Concordium-contracts-common` needs to be a dependency otherwise we cannot expose the `derive-serde` feature.

Applied your suggestion of using the wildcard character.

trackAndTrace/smart-contract/Cargo.toml
```diff
@@ -8,12 +8,14 @@ description = "A smart contract for tracking and tracing items on a supply chain
 
 [features]
 default = ["std", "wee_alloc"]
-std = ["concordium-std/std", "concordium-cis2/std"]
+std = ["concordium-std/std"]
 wee_alloc = ["concordium-std/wee_alloc"]
+derive-serde = ["concordium-contracts-common/derive-serde", "serde"]
```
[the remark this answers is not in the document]
DOBEN: We also have a dependency called `serde` and this causes an error:
```
    Caused by:
         optional dependency `serde` is not included in any feature
         Make sure that `dep:serde` is included in one of features in the [features] table.

```

trackAndTrace/indexer/README.md
```diff
@@ -0,0 +1,58 @@
...
+```console
+cargo build --release
+```
+
+This will produce a single binary `indexer` in `target/release` directory.
+
+## Start the indexer
+
+```console
+cargo run -- --node https://grpc.testnet.concordium.com:20000 --start 2024-01-28T10:12:30Z --contract "<7835,0>"
```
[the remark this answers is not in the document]
DOBEN: I use the block_height now. Since that value is easily obtained from `CCDScan` by copy-paste.

trackAndTrace/indexer/src/main.rs
```diff
@@ -0,0 +1,476 @@
...
+    let settings = db
+        .get_settings()
+        .await
+        .context("Could not get settings from database")?;
+
+    anyhow::ensure!(
+        settings.contract_address == app.contract_address,
+        "Contract address does not match the contract address found in the database"
+    );
+
```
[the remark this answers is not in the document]
DOBEN: The above check in the code, just ensured that we have correctly written the contract_address to the database.

Bargsteen (approved):
Nice work 👍

trackAndTrace/indexer/Cargo.toml
```diff
@@ -0,0 +1,23 @@
...
+tokio = { version = "1.35", features = ["rt-multi-thread", "macros", "sync", "signal"] }
+tonic = { version = "0.10", features = ["tls", "tls-roots"] }
+tokio-postgres = { version = "0.7", features = [
+  "with-serde_json-1",
+  "with-chrono-0_4",
+] }
+deadpool-postgres = "0.11"
+thiserror = "1.0"
+tracing = "0.1.40"
+tracing-subscriber = "0.3.18"
```
abizjak, line 22:
```suggestion
tracing-subscriber = "0.3"
```

trackAndTrace/indexer/Cargo.toml
```diff
@@ -0,0 +1,23 @@
...
+serde = { version = "1.0", features = ["derive"] }
+tokio = { version = "1.35", features = ["rt-multi-thread", "macros", "sync", "signal"] }
+tonic = { version = "0.10", features = ["tls", "tls-roots"] }
+tokio-postgres = { version = "0.7", features = [
+  "with-serde_json-1",
+  "with-chrono-0_4",
+] }
+deadpool-postgres = "0.11"
+thiserror = "1.0"
+tracing = "0.1.40"
```
abizjak, line 21:
```suggestion
tracing = "0.1"
```

trackAndTrace/indexer/Cargo.toml
```diff
@@ -0,0 +1,23 @@
...
+tracing-subscriber = "0.3.18"
+futures = "0.3.30"
```
abizjak, line 23:
```suggestion
futures = "0.3"
```

trackAndTrace/indexer/README.md
```diff
@@ -0,0 +1,67 @@
...
+
+All monitored events in a block are atomically added in one database transaction to postgres. This ensures a simple recovery process since we always process the complete block or roll back the database to the beginning of the block.
+
+In case the indexer stops, after fixing the cause of the failure the indexer can be re-started by providing as `--start` option the `latest_processed_block_height + 1`. You can figure out the latest processed block height from the logs of the indexer. Alternatively, you can get the `latest_processed_block_height_with_monitored_event` by taking the rows with the largest id in the `item_status_changed_events` and the `item_created_events` tables. Alternatively, you can re-start the indexer by providing as `--start` option the `latest_processed_block_height_with_monitored_event + 1`. You should ensure to never re-start the indexer from a block where it processes a block with monitored events a second time, since you will end up with duplicated rows in the database.
```
abizjak, line 9: I don't think we should be having this --start option. THis should be automatically discovered like it is for the transaction logger, and other indexers we have.

trackAndTrace/indexer/resources/schema.sql
```diff
@@ -0,0 +1,62 @@
...
+  contract_index INT8 NOT NULL,
+  -- Contract subindex. This will be set the frist time the indexer is started. 
+  -- Re-starting the indexer will check if its settings are compatible will 
+  -- the stored indexer setting to prevent corrupting the database.
+  contract_subindex INT8 NOT NULL,
+  -- The genesis block hash as querried from the node. This will be set the frist time the indexer is started. 
+  -- Re-starting the indexer will check if its settings are compatible will 
+  -- the stored indexer setting to prevent corrupting the database.
+  genesis_block_hash BYTEA NOT NULL
+);
```
abizjak, line 17: The indexer should store the last processed heigh or block, whatever is easiest.

So that restarting it requires no manual user intervention. This is what we want from all indexers. It is less error prone.

trackAndTrace/indexer/resources/schema.sql
abizjak: As I wrote in the comment above, the "settings" or another table if you want should store the "last processed block". So that restarting requires no user intervention.

If we ask the user to do it they will do it wrong, and we might skip events, or risk duplicates (unless we have enough database rules).

trackAndTrace/indexer/src/main.rs
```diff
@@ -0,0 +1,330 @@
...
+        if !contract_update_infos.is_empty() {
+            // Begin the transaction
+            let db_transaction = db
+                .client
+                .transaction()
+                .await
+                .context("Failed to build DB transaction")?;
+
+            for tx in contract_update_infos {
```
abizjak, line 232: The Rust SDK 4.2.0 is published now, so you might want to use that  to simplify this loop.

abizjak (commented):
Looks pretty good.

But I feel very strongly that the indexer should be able to restart without user intervention. This allows for maintenance for example.

trackAndTrace/indexer/resources/schema.sql
abizjak: Since you are processing blocks atomically you can, in the same DB transaction, also update this height. That will ensure consistency.

trackAndTrace/indexer/src/main.rs
```diff
@@ -0,0 +1,364 @@
...
+            "Encountered error {error}");
+
+        // Get new database connection from the pool
+        match self
+            .db_pool
+            .get()
+            .await
+            .context("Failed to get new database connection from pool")
+        {
+            Ok(_) => Ok(true),
```
abizjak, line 220: I don't think this will work. You want to set the active connection to this one, so
```
Ok(conn) => {
    self.db = conn;
}
```

trackAndTrace/indexer/src/main.rs
```diff
@@ -0,0 +1,364 @@
...
+                            .await
+                            .context("Failed to prepare item_created_event transaction")?;
+
+                        db_transaction
+                            .execute(&statement, &params)
+                            .await
+                            .context("Failed to execute item_created_event transaction")?;
+
+                        tracing::debug!(
+                            target:"ccd_event_processor",
```
abizjak, line 179: I'd recommend not setting an explicit target here so that it's clear that logs are comming from this module instead.

trackAndTrace/indexer/src/main.rs
```diff
@@ -0,0 +1,364 @@
...
+    log_level:        tracing_subscriber::filter::LevelFilter,
+}
+
+/// A handler for storing monitored events in the database. This implements
+/// the `indexer::ProcessEvent` trait to store events in the database.
+struct StoreEvents {
+    /// An active database connection to the postgres database.
+    db:      Database,
+    /// A database pool used for reconnects.
+    db_pool: DatabasePool,
```
abizjak, line 69: I think if you want to use a DatabasePool you should not keep a connection here as well, even though it is not wrong to do so.

I'd recommend just doing

```
let conn = db_pool.get().await?;
let db_transaction = conn.client.transaction().await...
```
inside the "process" method.

trackAndTrace/indexer/src/main.rs
```diff
@@ -0,0 +1,364 @@
...
+            target:"ccd_event_processor",
+            "Encountered error {error}");
+
+        // Get new database connection from the pool
+        match self
+            .db_pool
+            .get()
+            .await
+            .context("Failed to get new database connection from pool")
+        {
```
abizjak, line 219: If you follow my suggestion above then the "on_failure" does not really have to do anything other than log.

abizjak (approved):

limemloh (approved):

PR #49, opened 2024-02-06
CI for linting, formatting, and building frontends + Docker testing

### LINKED ISSUE #3: Set up automated testing
Since this repository is meant to contain examples it should be exemplary.

It should have tests to make sure it builds, that formatting is as dictated, and that tests pass.

lassemoldrup:
## Purpose

Closes #3.

## Changes

Adds CI for linting, formatting, and building example frontends.

## Checklist

      hard-to-understand areas.

### DIFF
--- a/sponsoredTransactionsAuction/frontend/src/components/App/App.tsx
+++ b/sponsoredTransactionsAuction/frontend/src/components/App/App.tsx
@@ -7,7 +7,6 @@ import {
     useConnection,
     useConnect,
     TESTNET,
-    useWalletConnectorSelector,
 } from '@concordium/react-components';
 import { AccountAddress, TransactionHash } from '@concordium/web-sdk';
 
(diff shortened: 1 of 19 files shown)


sponsoredTransactionsAuction/frontend/src/components/App/App.tsx
```diff
@@ -110,7 +110,7 @@ export default function App(props: WalletConnectionProps) {
 
     useEffect(() => {
         select();
-    }, []);
+    }, [select]);
```
DOBEN, line 113: Opening the front-end in firefox now constantly reloads/refreshes the page. A hot-fix would be to just keep the dependency array empty and disable the `eslint` rule here. We only want the `useEffect` hook to be executed once. Better would be of course to dig deeper if we can get this to work.

Opening the front-end in Chrome seems to work with the above changes so.

bisgardo: Both Firefox and Chrome exhibit the same behavior. The observed difference is caused by the browser wallet being installed in one and not the other: The "loading connector" message (which is what's flickering) only appears until the extension is loaded. So if you disable the wallet extension in Chrome, you'll see the same thing as Firefox.

The `select` function is intended to be triggered by a UI action (like clicking the "Use Browser Wallet" button in the wCCD dapp). Invoking it changes it's behavior to "deselect" the next time it's called (with a new function reference). Calling it every time the reference changes therefore causes the infinite select/deselect cycle observed.

I think in this case the correct solution might be to not use `useWalletConnectorSelector` becuase only the BW connector type is being used anyway. You could just call `setActiveConnectorType(BROWSER_WALLET)` directly and use `connection !== null` for `isConnected`.

If desired/required, it should be safe (but unnecessary and possibly confusing) to add the function to the dependency list because it refers to a fixed class method. I.e. I believe this should work:

```typescript
    useEffect(() => {
        setActiveConnectorType(BROWSER_WALLET);
    }, [setActiveConnectorType]);
```

Alternatively I also find the existing impl with the empty dependency array (suppressing the liner rule) to be acceptable.

DOBEN (approved):
Nice work.

Issue #50, opened 2024-02-07
Rust event indexer for the track-and-trace contract


DOBEN:
**Task description**

The task is to develop an example implementation of an event indexer written in Rust. 

The events from the `track-and-trace` smart contract will be indexed to serve the historical item statuses to the front end. The indexer starts processing old events and then listens for new events that are coming in as the blockchain progresses. The indexer should be configurable to start indexing at a specific time (usually the time around when the smart contract was deployed). The indexer should store `checkpoints` (`block_height`, `transaction_hash`, and `event_index`) of the current event being processed to aid in debugging and to aid in re-starting the index from the persistent database in case the indexer crashed unintentionally.

The indexer runs a PostgreSQL database at the backend.

There are three tables in the database:
- `setting` for global settings and the checkpoints.
- `item_created_events` for all the item-created events logged by the contract.
- `item_status_changed_events` for all the item-status-changed events logged by the contract.

Smart contracts:
https://github.com/Concordium/concordium-rust-smart-contracts/issues/387#issue-2097981242
https://github.com/Concordium/concordium-dapp-examples/tree/main/trackAndTrace

**Sub-tasks**
- [x] Develop an indexer to process events from the blockchain and input them into the PostgreSQL database.
- [x] Develop an http server that provides two entrypoints for the front end to query the database:
      - [x] `GetItemStatusChangedEvents` by `item_id` returns a list of all status-changed events for specified `item_id` or an empty list if the item does not exist.
      - [x] `GetItemCreatedEvent` by `item_id` returns the event when the item was created or `None` if the item does not exist.

*Add label for component and priority.*

PR #57, opened 2024-02-26
Server track and trace

DOBEN:
## Purpose

closes https://github.com/Concordium/concordium-dapp-examples/issues/50

 Develops a server that provides two endpoints for the front end to query the database.
 
 Related to: https://github.com/Concordium/concordium-dapp-examples/pull/54

## Changes

- Add server.
- Add `GetItemStatusChangedEvents` endpoint by `item_id` returns a vector of all status-changed events for specified `item_id` or an empty vector if the item does not exist.
- Add `GetItemCreatedEvent` endpoint by `item_id` returns the event when the item was created or None if the item does not exist.


sponsoredTransactionsAuction/backend/src/main.rs
```diff
@@ -204,10 +204,19 @@ async fn main() -> anyhow::Result<()> {
...
-        .with_graceful_shutdown(shutdown_signal)
-        .await?;
+
+    // Create the server.
+    let server = axum::Server::bind(&socket).serve(router.into_make_service());
+
+    // Wait for either the server to complete or a shutdown signal to be received.
+    tokio::select! {
+      _ = server => {
+          println!("Server has shut down gracefully");
```
abizjak, line 214: You should not `print`, you should use `tracing::info`.

sponsoredTransactionsAuction/backend/src/main.rs
```diff
@@ -204,10 +204,19 @@ async fn main() -> anyhow::Result<()> {
...
+      }
+      _ = shutdown_signal => {
+          println!("Received shutdown signal. Shutting down server gracefully...");
+      }
```
abizjak, line 218: I don't understand why you removed the `with_graceful_shutdown`. That is supposed to ensure that existing clients are served before the server shuts down.

The way you have it now means the server is just killed very ungracefully.

trackAndTrace/indexer/src/bin/server.rs
```diff
@@ -0,0 +1,287 @@
...
+}
+
+impl axum::response::IntoResponse for ServerError {
+    fn into_response(self) -> axum::response::Response {
+        let r = match self {
+            ServerError::DatabaseErrorPostgres(error) => {
+                tracing::error!("Internal error: {error}.");
+                (
+                    StatusCode::INTERNAL_SERVER_ERROR,
+                    Json(format!("{}", error)),
```
abizjak, line 53: We don't want to be returning internal errors to clients. It might accidentally leak sensitive information.

trackAndTrace/indexer/src/bin/server.rs
```diff
@@ -0,0 +1,287 @@
...
+                )
+            }
+            ServerError::DatabaseErrorTypeConversion(error) => {
+                tracing::error!("Internal error: {error}.");
+                (StatusCode::INTERNAL_SERVER_ERROR, Json(error.to_string()))
```
abizjak, line 58: We don't want to be returning internal errors to clients. It might accidentally leak sensitive information.

trackAndTrace/indexer/src/bin/server.rs
```diff
@@ -0,0 +1,287 @@
...
+            }
+            ServerError::DatabaseErrorConfiguration(error) => {
+                tracing::error!("Internal error: {error}.");
+                (
+                    StatusCode::INTERNAL_SERVER_ERROR,
+                    Json(format!("{}", error)),
```
abizjak, line 64: We don't want to be returning internal errors to clients. It might accidentally leak sensitive information.

trackAndTrace/indexer/src/bin/server.rs
```diff
@@ -0,0 +1,287 @@
...
+                )
+            }
+            error => {
```
abizjak, line 67: Please make all the cases explicit for better maintainability.

trackAndTrace/indexer/src/bin/server.rs
```diff
@@ -0,0 +1,287 @@
...
+        r.into_response()
+    }
+}
+
+/// Command line configuration of the application.
+#[derive(Debug, clap::Parser)]
+#[command(author, version, about)]
+struct Args {
+    #[clap(
+        long = "port",
```
abizjak, line 81: This should not be called port, it is the entire listen-address.

trackAndTrace/indexer/src/bin/server.rs
```diff
@@ -0,0 +1,287 @@
...
+    let db_pool = DatabasePool::create(app.db_connection.clone(), 1, true)
+        .await
+        .context("Could not create database pool")?;
+
+    let state = Server { db_pool };
+
+    tracing::info!("Starting server...");
+
+    // Render `index.html` file and `assets` folder.
+    let index_template = fs::read_to_string(app.frontend_assets.join("index.html"))
```
abizjak, line 141: Why is this called a "template" when it's just a file?

trackAndTrace/indexer/src/bin/server.rs
```diff
@@ -0,0 +1,287 @@
...
+
+    // Wait for either the server to complete or a shutdown signal to be received.
+    tokio::select! {
+      _ = server => {
+          println!("Server has shut down gracefully");
+      }
+      _ = shutdown_signal => {
+          println!("Received shutdown signal. Shutting down server gracefully...");
+      }
+    }
```
abizjak, line 175: Please use with_graceful_shutdown for the server, not this.

trackAndTrace/indexer/src/bin/server.rs
```diff
@@ -0,0 +1,287 @@
...
+/// Struct returned by the `getItemStatusChangedEvents` endpoint. It returns a
+/// vector of ItemStatusChangedEvents from the database if present.
+#[derive(serde::Serialize)]
+struct StoredItemStatusChangedEventsReturnValue {
+    data: Vec<StoredItemStatusChangedEvent>,
+}
+
+/// Handles the `getItemStatusChangedEvents` endpoint, returning a vector of
+/// ItemStatusChangedEvents from the database if present.
+#[tracing::instrument(level = "info")]
```
abizjak, line 245: This will print the entire state in the log. We don't need that.

trackAndTrace/indexer/src/bin/server.rs
```diff
@@ -0,0 +1,287 @@
...
+/// itemCreatedEvent from the database if present.
+#[derive(serde::Serialize)]
+struct StoredItemCreatedEventReturnValue {
+    data: Option<StoredItemCreatedEvent>,
+}
+
+/// Handles the `health` endpoint, returning the itemCreatedEvent from the
+/// database if present.
+#[tracing::instrument(level = "info")]
+async fn get_item_created_event(
```
abizjak, line 274: Same as above, we don't want to print all the parameters.

trackAndTrace/indexer/src/bin/server.rs
```diff
@@ -0,0 +1,287 @@
...
+    request: Result<Json<u64>, JsonRejection>,
+) -> Result<Json<StoredItemStatusChangedEventsReturnValue>, ServerError> {
+    let db = state.db_pool.get().await?;
+
+    let Json(item_id) = request?;
+
+    // We hardcode the pagination here for simplicity for this demo dApp.
+    let database_result = db
+        .get_item_status_changed_events_submissions(item_id, 30, 0)
+        .await?;
```
abizjak, line 257: This is not really acceptable. It is just incorrect to always go from 0. So we have to add additional (optional) parameters to the request, limit and offset.

And guard that the limit is not too large.

abizjak (changes requested):

trackAndTrace/indexer/src/bin/server.rs
```diff
@@ -0,0 +1,301 @@
...
+    State(state): State<Server>,
+    request: Result<Json<GetItemstatusChangedEventsParam>, JsonRejection>,
+) -> Result<Json<StoredItemStatusChangedEventsReturnValue>, ServerError> {
+    let db = state.db_pool.get().await?;
+
+    let Json(param) = request?;
+
+    if param.limit > MAX_REQUEST_LIMIT {
+        return Err(ServerError::MaxRequestLimit(MAX_REQUEST_LIMIT));
+    }
```
abizjak, line 268: In the wallet-proxy we just take the `min(MAX_REQUEST_LIMIT, param.limit)`. 

Both options are fine with me, I'm just wondering why you chose to reject the request.

trackAndTrace/indexer/src/bin/server.rs
```diff
@@ -0,0 +1,301 @@
...
+pub enum ServerError {
+    #[error("Database error from postgres: {0}")]
+    DatabaseErrorPostgres(tokio_postgres::Error),
+    #[error("Database error in type conversion: {0}")]
+    DatabaseErrorTypeConversion(String),
+    #[error("Database error in configuration: {0}")]
+    DatabaseErrorConfiguration(anyhow::Error),
+    #[error("Failed to extract json object: {0}")]
+    JsonRejection(#[from] JsonRejection),
+    #[error("The requested events to the database where above the limit {0}")]
```
abizjak, line 36:
```suggestion
    #[error("The requested events to the database were above the limit {0}")]
```

abizjak (approved):

trackAndTrace/indexer/src/bin/server.rs
DOBEN: My thinking was along the way, that it is easier to catch from the front end that pagination needs to be implemented if the request reverts rather than returning silently fewer items than requested.

PR #59, opened 2024-02-29
Add generic sponsored transaction service

Bargsteen:
## Purpose

Add a generic sponsored transaction service that works with any CIS-3 contract.

There is a single todo about how permissive CORS should be. Right now it allows all connections.

## Changes

- Add a new service.
- Change Rust version in the relevant CI job from `1.72` to `1.74`

## Checklist

      hard-to-understand areas.


trackAndTrace/backend/src/main.rs
```diff
@@ -0,0 +1,441 @@
...
+
+    let rate_limit = rate_limits.entry(alias_account_0).or_insert_with(|| 0u8);
+
+    if *rate_limit >= RATE_LIMIT_PER_ACCOUNT {
+        tracing::warn!("Rate limit for account {:#?} reached.", request.signer);
+
+        return Err(ServerError::RateLimitError);
+    }
+
+    *rate_limit += 1;
```
DOBEN, line 371: This way of implementing the `rate_limit` might not be very suitable for the `track_and_trace` project. Currently, we only allow every whitelisted account to submit up to 30 transactions and then no more. 

Since we suspect a whitelisted account to submit a lot of transactions over time. The rate limit should be implemented over time. E.g. no more than 30 transactions per hour.

trackAndTrace/backend/src/types.rs
```diff
@@ -0,0 +1,233 @@
...
+                    Json(format!("{}", error)),
+                )
+            }
+            ServerError::SubmitSponsoredTransactionError(error) => {
+                tracing::error!("Internal error: {error}.");
+                (
+                    StatusCode::INTERNAL_SERVER_ERROR,
+                    Json(format!("{}", error)),
+                )
+            }
```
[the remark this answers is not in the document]
DOBEN: We had this discussion that we don't want to forward internal errors to the front end to avoid liking information in general. On the other hand, if the internal error comes from a reverting smart contract during the simulation, it is nice to have the error code at the front-end to debug it.

DOBEN:
```There is a single todo about how permissive CORS should be. Right now it allows all connections.```
I suggest we do this via an environmental variable, so the allowed origin can be specified when starting the backend server.

File: trackAndTrace/backend/src/types.rs [hunk 3]
```diff
@@ -0,0 +1,233 @@
...
+/// Either accounts or contracts.
+///
+/// The expected argument format is
+/// "any", "Any", or "ANY" for `Self::Any`, and a white-space-separated set of
+/// the entities for the `Self::LimitedTo` variant.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub enum AllowedEntities<T> {
+    /// Any entities allowed to use or be used by the service. This should only
+    /// be used if other authorizations schemes are in place. Otherwise the
+    /// sponsor account can easily be drained of funds. Especially for the
```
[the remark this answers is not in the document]
Bargsteen: I deemed it more important to limit the accounts than the contracts, but that is contingent on a high level of trust between the sponsor and whitelisted accounts. Since that isn't guaranteed, I'll just remove the sentence.

trackAndTrace/backend/README.md
```diff
@@ -0,0 +1,116 @@
...
+  - Equivalent environment flag: `CCD_SPONSORED_TRX_SERVICE_LISTEN_ADDRESS`.
+- `log-level` maximum log level (defaults to `debug` if not given).
+  - Equivalent environment flag: `CCD_SPONSORED_TRX_SERVICE_LOG_LEVEL`.
+- `request-timeout` the timeout for server and node requests in milliseconds. Defaults to `10000`.
+  - Equivalent environment flag: `CCD_SPONSORED_TRX_SERVICE_REQUEST_TIMEOUT`.
+- `account` the path to a file which contains the key credentials for the sponsor account.
+  - Equivalent environment flag: `CCD_SPONSORED_TRX_SERVICE_PRIVATE_KEY_FILE`.
+- `allowed-accounts` The accounts allowed to submit transactions. Either 'any', if you have a custom authentication scheme in front of the service OR a space-separated list of account addresses.
+  - Equivalent environment flag: `CCD_SPONSORED_TRX_SERVICE_ALLOWED_ACCOUNTS`.
+- `allowed-contracts` The contracts allowed to be used by the service. Either 'any' OR a space-separated list of contract addresses in the format `<123,0>`.
```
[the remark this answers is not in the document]
Bargsteen: Duplicates are allowed but since it uses a set inside, they will be dedupped. We could change the `list` to `set` in the sentence, but I think `list` is less ambiguous in terms of the format. Whether duplicates are allowed or not is really not that relevant, as the presence of an entry of the list is clear even if duplicated.

trackAndTrace/backend/Cargo.toml
```diff
@@ -0,0 +1,37 @@
...
+tokio = { version = "1", features = ["full"] }
+tonic = {version = "0.10", features = ["tls", "tls-roots"]} # Use system trust roots.
+serde = { version = "1.0", features = ["derive"] }
+serde_json = "1.0"
+clap = { version = "4", features = ["derive", "env"] }
+anyhow = "1.0"
+chrono = "0.4.19"
+thiserror = "1"
+rand = "0.8"
+uuid = { version = "1.2.2", features = ["v4", "serde"] }
```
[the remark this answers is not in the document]
Bargsteen: Eagle eyes, nicely spotted :D

trackAndTrace/backend/Cargo.toml
```diff
@@ -0,0 +1,37 @@
...
+hex = "0.4.3"
+concordium-rust-sdk = "4.2.0"
+concordium-contracts-common = { version = "*", features = ["derive-serde"] }
```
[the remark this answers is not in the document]
Bargsteen: For the derive-serde feature which cannot be enabled through the rust SDK. I'll add a comment about it.

trackAndTrace/sponsored-transaction-service/src/main.rs
```diff
@@ -0,0 +1,438 @@
...
+
+/// Handle a request for a sponsored transaction.
+pub async fn handle_transaction(
+    State(mut state): State<Server>,
+    request: Result<Json<InputParams>, JsonRejection>,
+) -> Result<Json<TransactionHash>, ServerError> {
+    let Json(request) = request?;
+
+    if !state.allowed_accounts.allowed(&request.signer) {
+        tracing::debug!("Account not allowed: {}", request.signer);
```
abizjak, line 214: Isn' this logging duplicated? The error handler also logs this because of bad request.

trackAndTrace/sponsored-transaction-service/src/main.rs
```diff
@@ -0,0 +1,438 @@
...
+        return Err(ServerError::AccountNotAllowed {
+            account: request.signer,
+        });
+    }
+    if !state.allowed_contracts.allowed(&request.contract_address) {
+        tracing::debug!("Contract not allowed: {}", request.contract_address);
```
abizjak, line 220: Same here.

trackAndTrace/sponsored-transaction-service/src/main.rs
```diff
@@ -0,0 +1,438 @@
...
+
+    let mut inner_signature_map = BTreeMap::new();
+    inner_signature_map.insert(0, Signature::Ed25519(SignatureEd25519(signature)));
+
+    let mut signature_map = BTreeMap::new();
+    signature_map.insert(0, CredentialSignatures {
+        sigs: inner_signature_map,
+    });
+
+    tracing::debug!("Create Parameter.");
```
abizjak, line 247: This logging is only really appropriate for `trace`. And ideally if you want to keep these you'd make a span that would identify what request they are a part of. https://docs.rs/tracing/latest/tracing/attr.instrument.html#adding-fields

trackAndTrace/sponsored-transaction-service/src/main.rs
```diff
@@ -0,0 +1,438 @@
...
+    };
+
+    // Transaction should expire after one hour.
+    let transaction_expiry = TransactionTime::hours_after(1);
+
+    // Get the current nonce for the backend wallet and lock it. This is necessary
+    // since it is possible that API requests come in parallel. The nonce is
+    // increased by 1 and its lock is released after the transaction is submitted to
+    // the blockchain.
+    let mut nonce = state.nonce.lock().await;
```
abizjak, line 332: And then send it using this builder https://docs.rs/concordium-rust-sdk/latest/concordium_rust_sdk/contract_client/type.ContractUpdateBuilder.html which is returned in case of a successful dry-run.

trackAndTrace/sponsored-transaction-service/src/main.rs
```diff
@@ -0,0 +1,438 @@
...
+    match state.node_client.send_block_item(&bi).await {
+        Ok(hash) => {
+            tracing::debug!("Submit transaction {} ...", hash);
+
+            *nonce = nonce.next();
+
+            Ok(hash.into())
+        }
+        Err(e) => {
+            tracing::warn!("SubmitSponsoredTransactionError {e}.");
```
abizjak, line 395: Isn't this going to be logged already by axum? The error handler has logs as far as I see.

trackAndTrace/sponsored-transaction-service/Cargo.toml
```diff
@@ -0,0 +1,36 @@
...
+serde = { version = "1.0", features = ["derive"] }
+serde_json = "1.0"
+clap = { version = "4", features = ["derive", "env"] }
+anyhow = "1.0"
+chrono = "0.4.19"
+thiserror = "1"
+rand = "0.8"
+hex = "0.4.3"
+concordium-rust-sdk = "4.2.0"
+concordium-contracts-common = { version = "*", features = ["derive-serde"] } # Added directly to enable the 'derive-serde' feature.
```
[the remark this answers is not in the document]
Bargsteen: You are right, it isn't needed. Perhaps I had the wrong or mismatched versions of something. Thanks!

trackAndTrace/sponsored-transaction-service/src/types.rs
```diff
@@ -0,0 +1,237 @@
...
+                (
+                    StatusCode::INTERNAL_SERVER_ERROR,
+                    Json(format!("{}", error)),
+                )
+            }
+            ServerError::SubmitSponsoredTransactionError(error) => {
+                tracing::error!("Internal error: {error}.");
+                (
+                    StatusCode::INTERNAL_SERVER_ERROR,
+                    Json(format!("{}", error)),
```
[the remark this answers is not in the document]
Bargsteen: Ok, I've changed it, so that we only return a static message in these cases instead of forwarding the internal error.

trackAndTrace/sponsored-transaction-service/src/main.rs
```diff
@@ -0,0 +1,438 @@
...
+        .layer(CorsLayer::permissive()) // TODO: Do we want to restrict CORS?
+        .layer(
+            tower_http::trace::TraceLayer::new_for_http()
+                .make_span_with(tower_http::trace::DefaultMakeSpan::new())
+                .on_response(tower_http::trace::DefaultOnResponse::new()),
+        )
+        .layer(tower_http::timeout::TimeoutLayer::new(
+            std::time::Duration::from_millis(app.request_timeout),
+        ))
+        .layer(tower_http::limit::RequestBodyLimitLayer::new(1_000_000)) // at most 1000kB of data.
```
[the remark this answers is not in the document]
Bargsteen: Good point. Changed to 70kb ;)

trackAndTrace/sponsored-transaction-service/src/main.rs
```diff
@@ -0,0 +1,438 @@
...
+            .with(log_filter)
+            .init();
+    }
+
+    let endpoint = if app
+        .endpoint
+        .uri()
+        .scheme()
+        // We compare on strings as they use different versions of the `http` crate, so the types cannot be compared.
+        .map_or(false, |x| x.as_str() == http::uri::Scheme::HTTPS.as_str())
```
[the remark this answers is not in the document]
Bargsteen: The Rust SDK uses `http = 0.2` but this service uses `http = 1.0`. I assume that is why I get
```
error[E0277]: can't compare `concordium_rust_sdk::v2::Scheme` with `http::uri::Scheme`
   --> trackAndTrace/sponsored-transaction-service/src/main.rs:123:31
    |
123 |         .map_or(false, |x| *x == http::uri::Scheme::HTTPS)
    |                               ^^ no implementation for `concordium_rust_sdk::v2::Scheme == http::uri::Scheme`
    |
    = help: the trait `PartialEq<http::uri::Scheme>` is not implemented for `concordium_rust_sdk::v2::Scheme`
    = help: the following other types implement trait `PartialEq<Rhs>`:
              <concordium_rust_sdk::v2::Scheme as PartialEq>
              <concordium_rust_sdk::v2::Scheme as PartialEq<str>>
```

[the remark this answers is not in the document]
Bargsteen: Ah, of course, thanks ;) That _does_ work.

trackAndTrace/sponsored-transaction-service/src/main.rs
```diff
@@ -0,0 +1,411 @@
...
+    });
+
+    let param: PermitParam = PermitParam {
+        message,
+        signature: AccountSignatures {
+            sigs: signature_map,
+        },
+        signer: request.signer,
+    };
+
```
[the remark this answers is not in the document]
Bargsteen: I did not catch that you suggested this.
The phrasing was unclear on https://github.com/Concordium/concordium-dapp-examples/pull/59#discussion_r1518973374. I actually thought it was a follow-up to your comment about tracing and spans, which I then dismissed as I simply removed the tracing.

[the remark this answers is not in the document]
Bargsteen: That made it much nicer, thanks! :D

trackAndTrace/sponsored-transaction-service/src/main.rs
```diff
@@ -0,0 +1,340 @@
...
+               .allow_methods([http::Method::POST]))
+        .layer(
+            tower_http::trace::TraceLayer::new_for_http()
+                .make_span_with(tower_http::trace::DefaultMakeSpan::new())
+                .on_response(tower_http::trace::DefaultOnResponse::new()),
+        )
+        .layer(tower_http::timeout::TimeoutLayer::new(
+            std::time::Duration::from_millis(app.request_timeout),
+        ))
+        .layer(tower_http::limit::RequestBodyLimitLayer::new(70_000)) // Allow at most 70kB of data. Enough for max parameter u16::MAX and the signatures etc.
```
abizjak, line 205: Is standard formatting set up? This seems like an excessively long line for that.

trackAndTrace/sponsored-transaction-service/src/main.rs
```diff
@@ -0,0 +1,340 @@
...
+    let param: PermitParam = PermitParam {
+        message,
+        signature: AccountSignatures {
+            sigs: signature_map,
+        },
+        signer: request.signer,
+    };
+
+    let parameter = smart_contracts::OwnedParameter::from_serial(&param)
+        .map_err(|_| ServerError::ParameterError)?;
```
abizjak, line 266: Any particular reason for constructing the parameter here and using the `_raw` method later on, rather than just using the method without `_raw`?

trackAndTrace/sponsored-transaction-service/src/main.rs
```diff
@@ -0,0 +1,340 @@
...
+    // since it is possible that API requests come in parallel. The nonce is
+    // increased by 1 and its lock is released after the transaction is submitted to
+    // the blockchain.
+    let mut nonce = state.nonce.lock().await;
+
+    // Reset the rate limits if an hour has passed since the last reset.
+    state.reset_rate_limits_if_expired().await;
+
+    // Check the rate limits for the account.
+    state.check_rate_limit(request.signer).await?;
```
abizjak, line 288: Shouldn't we check this before the dry_run?

abizjak (approved):

trackAndTrace/sponsored-transaction-service/src/main.rs
Bargsteen: No, fixed.

trackAndTrace/sponsored-transaction-service/src/main.rs
Bargsteen: No, we didn't have a `rustfmt.toml` file, but I copied one over from the rust SDK and ran `rust fmt` on nightly. It didn't fix the line though. Perhaps it is because it is in the middle of a chain of calls. I split the line manually.

trackAndTrace/sponsored-transaction-service/src/main.rs
Bargsteen: I'll merge this now as that makes it easier for my docker PR, but I can make a new PR if we want to make changes to the rate limits.

abizjak: I think we should check the limit before, not that we should increase it.

PR #60, opened 2024-03-01
Add initial generalized sponsored transaction backend

DOBEN:
## Purpose

Add sponsored transaction backend. The backend accepts a trusted list of whitelisted accounts. Only these accounts can send sponsored transactions via the backend. The backend is generalized, so it can be easily adapted to serve any contract implementing the `CIS3` standard. As such, contract address, name, payload are sent from the front end to the backend and are not hardcoded at the backend.

TODO: It would be nice to import `PermitMessage` and `PermitParam` from the smart contract. Currently, we miss `serde::Deserialize` and `serde:Serialize` traits on the `AccountSignature` to do that. I plan to do this next.

## Changes

Add sponsored transaction backend.


DOBEN:
Closing it, because we decided to move the sponsored transaction backend to a separate microservice in this PR.
https://github.com/Concordium/concordium-dapp-examples/pull/59

PR #64, opened 2024-03-15
Add docker and docker compose files to track and trace project

Bargsteen:
## Purpose

Create docker files for the whole track and trace project.

## Changes

- Add dockerfiles for each part of the project
- Add a docker compose file for running the whole project
- Add a `.env` file for configuring the most common options
- Remove default environment variables in `./frontend/package.json` since they took precedence over values from the docker files.
- Update READMEs

## Checklist

      hard-to-understand areas.


trackAndTrace/docker-compose.yml
```diff
@@ -0,0 +1,67 @@
+version: '3.8'
```
DOBEN, line 1:
```suggestion
version: '1.0'
```

trackAndTrace/docker-compose.yml
```diff
@@ -0,0 +1,67 @@
...
+
+# This docker compose setup has two important configuration options, which reside in the `.env` file.
```
DOBEN, line 3:
```suggestion
# This docker compose setup has three important configuration options, which reside in the `.env` file.
```

trackAndTrace/dockerfiles/server.Dockerfile
```diff
@@ -0,0 +1,39 @@
...
+# Forward the build args to `yarn build` via environment variables.
+# We do not use the docker `ENV` as those are only used for the final `CMD`.
+RUN TRACK_AND_TRACE_CONTRACT_INDEX=${TRACK_AND_TRACE_CONTRACT_INDEX} \
+SPONSORED_TRANSACTION_BACKEND_BASE_URL=${SPONSORED_TRANSACTION_BACKEND_BASE_URL} \
+yarn build
+
+# Build server
+FROM ${RUST_IMAGE} AS server
+COPY --from=frontend ./frontend/dist ./frontend/dist
+COPY ./smart-contract ./smart-contract
```
DOBEN, line 30: Do we need this?

DOBEN (approved):
Nice work.

trackAndTrace/.env
```diff
@@ -0,0 +1,3 @@
+TRACK_AND_TRACE_CONTRACT_INDEX=8291
```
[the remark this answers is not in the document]
Bargsteen: Ah, my mistake :p

DOBEN:
Something we need as well: support connecting to the node via TLS in docker images

https://github.com/Concordium/concordium-governance-committee-voting/pull/107/files

DOBEN:
Something we need as well: support connecting to the node via TLS in docker images

https://github.com/Concordium/concordium-governance-committee-voting/pull/107/files

trackAndTrace/dockerfiles/indexer.Dockerfile
```diff
@@ -0,0 +1,16 @@
...
+COPY ./smart-contract ./smart-contract
+WORKDIR /indexer
+COPY ./indexer/Cargo.toml ./Cargo.toml
+COPY ./indexer/Cargo.lock ./Cargo.lock
+COPY ./indexer/src ./src
+COPY ./indexer/resources ./resources
+RUN cargo build --release
+
+# Run indexer
+CMD ["cargo", "run", "--release", "--bin", "indexer"]
```
abizjak, line 16: I don't think this is right. You want to build the binary once, and then copy just the artifact into the final image.

So you want a multi-stage image
```suggestion
ARG RUST_IMAGE=rust:1.74-buster

# Build indexer
FROM ${RUST_IMAGE} as build
COPY ./smart-contract ./smart-contract
WORKDIR /indexer
COPY ./indexer/Cargo.toml ./Cargo.toml
COPY ./indexer/Cargo.lock ./Cargo.lock
COPY ./indexer/src ./src
COPY ./indexer/resources ./resources
RUN cargo build --release

FROM debian:buster

# In order to use TLS when connecting to the node we need certificates.
RUN apt-get update && apt-get install -y ca-certificates

COPY --from=build /indexer/target/release/indexer /indexer

# Run indexer
CMD ["indexer"]
```

trackAndTrace/dockerfiles/server.Dockerfile
```diff
@@ -0,0 +1,39 @@
...
+WORKDIR /server
+COPY ./indexer/Cargo.toml ./Cargo.toml
+COPY ./indexer/Cargo.lock ./Cargo.lock
+COPY ./indexer/src ./src
+COPY ./indexer/resources ./resources
+RUN cargo build --release
+
+# Run server
+CMD ["cargo", "run", "--release", "--bin", "server"]
```
abizjak, line 39: Same here.

trackAndTrace/dockerfiles/sponsored-transaction-service.Dockerfile
```diff
@@ -0,0 +1,13 @@
...
+
+# Build service
+FROM ${RUST_IMAGE}
+COPY ./sponsored-transaction-service/Cargo.toml ./Cargo.toml
+COPY ./sponsored-transaction-service/Cargo.lock ./Cargo.lock
+COPY ./sponsored-transaction-service/src ./src
+RUN cargo build --release
+
+# Run service
+CMD ["cargo", "run", "--release"]
```
abizjak, line 13: And here.

trackAndTrace/dockerfiles/indexer.Dockerfile
Bargsteen: Ah, that makes sense. I used `bookwork` instead of `buster` though, as `buster` no longer receives security updates.

.github/workflows/yarn-build-fmt-lint.yaml
```diff
@@ -129,6 +129,16 @@ jobs:
         working-directory: ${{ matrix.packages }}
         run: yarn install
 
-      - name: Build
+      - name: Build trackAndTrace package
+        if: ${{ matrix.packages == 'trackAndTrace/frontend' }}
         working-directory: ${{ matrix.packages }}
+        env:
+          TRACK_AND_TRACE_CONTRACT_INDEX: 8219
```
abizjak, line 136: Maybe add a comment that this is just some dummy value for the build to succeed?

trackAndTrace/.env
```diff
@@ -0,0 +1,4 @@
+TRACK_AND_TRACE_CONTRACT_ADDRESS=<8351,0>
+TRACK_AND_TRACE_PRIVATE_KEY_FILE=/private-keys/my-account.export
+TRACK_AND_TRACE_NODE=https://grpc.testnet.concordium.com:20000
+TRACK_AND_TRACE_NETWORK=testnet
```
abizjak, line 4: I'm not a big fan of these .env files in repositories. Especially ones such as this where you have links to private keys in them (which users are presumably not required to put in that location).

Would it not be better to be explicit about these configuration options when asking users to run things?

trackAndTrace/docker-compose.yml
```diff
@@ -0,0 +1,69 @@
...
+      dockerfile: ./dockerfiles/sponsored-transaction-service.Dockerfile
+    restart: always
+    ports:
+      - 8000:8000
+    volumes:
+      - ./private-keys:/private-keys
+    environment:
+      CCD_SPONSORED_TRANSACTION_SERVICE_ALLOWED_ACCOUNTS: Any
+      CCD_SPONSORED_TRANSACTION_SERVICE_ALLOWED_CONTRACTS: ${TRACK_AND_TRACE_CONTRACT_ADDRESS}
+      CCD_SPONSORED_TRANSACTION_SERVICE_PRIVATE_KEY_FILE: ${TRACK_AND_TRACE_PRIVATE_KEY_FILE}
```
abizjak, line 23: Presumably this refers to the location inside the built docker image? In which case, why is it not hardcoded here instead of in the .env file?

trackAndTrace/docker-compose.yml
```diff
@@ -0,0 +1,69 @@
...
+    build:
+      context: .
+      dockerfile: ./dockerfiles/sponsored-transaction-service.Dockerfile
+    restart: always
+    ports:
+      - 8000:8000
+    volumes:
+      - ./private-keys:/private-keys
+    environment:
+      CCD_SPONSORED_TRANSACTION_SERVICE_ALLOWED_ACCOUNTS: Any
```
abizjak, line 21: Maybe clarify that this should usually be limited.

trackAndTrace/indexer/src/bin/server.rs
```diff
@@ -149,18 +200,30 @@ async fn main() -> anyhow::Result<()> {
...
         .route("/api/getItemCreatedEvent", post(get_item_created_event))
         .route("/health", get(health))
         .nest_service("/assets", serve_dir_service)
-        .fallback(get(|| async { Html(index_file) }))
+        .fallback(get(|| async { Html(index_html) }))
         .with_state(state)
+        // .layer(CorsLayer::permissive())
+               // .allow_origin(AllowOrigin::mirror_request())
+               // .allow_headers([http::header::CONTENT_TYPE])
+               // .allow_methods([http::Method::POST, http::Method::GET]))
```
abizjak, line 226: Is there a reason for commenting this out?

trackAndTrace/docker-compose.yml
```diff
@@ -0,0 +1,69 @@
...
+services:
+  sponsored-transaction-service:
+    build:
+      context: .
+      dockerfile: ./dockerfiles/sponsored-transaction-service.Dockerfile
+    restart: always
+    ports:
+      - 8000:8000
+    volumes:
+      - ./private-keys:/private-keys
```
abizjak, line 19: Maybe add a comment that this is something they might want to change.

trackAndTrace/dockerfiles/server.Dockerfile
```diff
@@ -0,0 +1,44 @@
...
+RUN yarn build
+
+# Build server
+FROM ${RUST_IMAGE} AS server
+COPY ./smart-contract ./smart-contract
+WORKDIR /server
+COPY ./indexer/Cargo.toml ./Cargo.toml
+COPY ./indexer/Cargo.lock ./Cargo.lock
+COPY ./indexer/src ./src
+COPY ./indexer/resources ./resources
```
abizjak, line 32: Why not just one
```
COPY ./indexer ./
```

You can add a .dockerignore file like this https://github.com/Concordium/concordium-governance-committee-voting/blob/main/.dockerignore to ignore build artifacts you don't want to copy in.

trackAndTrace/frontend/README.md
```diff
@@ -18,52 +18,51 @@ yarn install
...
+To start a development environment make sure to first generate the smart contract clients, then run the following from the `frontend` directory:
 
 ```bash
 yarn dev
 ```
 
 This will launch a development server with hot module replacement enabled.
 
-If you want to set up the whole project together with running the backend services so that the 3 endpoints work. Follow the `README.md` files at the backend services. In detail, the front end files need to be served through the [backend server](https://github.com/Concordium/concordium-dapp-examples/tree/main/trackAndTrace/indexer) and the `SPONSORED_TRANSACTION_BACKEND_BASE_URL` needs to be set correctly and its backend service needs to be up and running.
+_Please note that calls to the backend server does not work in the development setup._
```
abizjak, line 39:
```suggestion
_Please note that calls to the backend server do not work in the development setup._
```

abizjak (approved):

trackAndTrace/indexer/src/bin/server.rs
```diff
@@ -111,16 +114,64 @@ struct Args {
...
+    #[clap(
+        long = "network",
+        default_value_t = concordium_rust_sdk::web3id::did::Network::Testnet,
+        help = "The network to connect users to (passed to frontend). Possible values: testnet, mainnet",
+        env = "CCD_SERVER_NETWORK",
+    )]
+    network: concordium_rust_sdk::web3id::did::Network,
+    /// The contract address of the election contract (passed to frontend).
+    #[clap(
+        long = "contract-address",
```
DOBEN, line 147: Can we add a default value here or change the read-me command to contain this flag:

```cargo run --bin server  -- --contract-address "<8219,0>"```

**Comment by DOBEN** [comment]
Why do we use `rust-bookworm` in the dockerfile? Just curious why we need that specific one.

**Comment by abizjak** [comment]
> Why do we use `rust-bookworm` in the dockerfile? Just curious why we need that specific one.

It's important that the base image that is used to run the binary and the one used to build it are compatible since otherwise some dependencies, like libc, might be different versions leading to failures.

**Review by DOBEN — APPROVED** [review]

File: trackAndTrace/indexer/src/bin/server.rs (hunk 16 above)
Comment (abizjak) [reply trackAndTrace/indexer/src/bin/server.rs:147]: I don't think it makes sense to have defaults for a contract. It's not something that has a meaningful default.

Comment (DOBEN) [reply trackAndTrace/indexer/src/bin/server.rs:147]: Please update the section: "## Configure the `server`` in the READ.me to include all the added additional options.

I suggest to add the READ.me with the following:
```
An example of running the service with basic settings and testnet node would be:

`
cargo run --bin server  -- --contract-address <YOUR_CONTRACT_ADDRESS>
`

An example to run the service with some filled in example settings would be:

`
cargo run --bin server  -- --contract-address "<8219,0>"
`
```

.github/workflows/yarn-build-fmt-lint.yaml
Bargsteen: Oh, these aren't needed anymore. Thanks for pointing them out.

trackAndTrace/docker-compose.yml
Bargsteen: I configured it such that you provide a path to the exact file, which is then mapped to a fixed location in the container.

trackAndTrace/indexer/src/bin/server.rs
Bargsteen: I added this while trying to get the complete frontend to work even when running `yarn dev` (along with the backend server). But there were some issues, and I don't think that particular feature is necessary. 
So I'll remove the cors layer again, as it isn't needed when the server also hosts the frontend.

trackAndTrace/docker-compose.yml
abizjak: Yes, my point is that the outside path is something that is reasonable to change, but the path inside the docker container is not really useful to change. It can just be fixed.

My comment was about elaborating on that.

Bargsteen: Ok, I believe the readme covers this sufficiently now.

PR #65, opened 2024-03-18
Euroe demo with walletconnect

orhoj:
## Purpose
Use dapp-libraries to connect to Browser Wallets and Mobile Wallets (through WalletConnect).

## Changes
- Now using dapp-libraries for maintaining the connection to a wallet.
- On the base URL the dApp will try to connect to a browser wallet, and on `/mobile` it will use WalletConnect. This is done to keep the webpage as simple as possible by avoiding a selector. This hopefully makes demoing cleaner.
- The dApp now requires the proof to be from the currently connected account in a wallet, so if the user decides to select a different account to provide the proof, then the dApp will reject them. Currently the dapp-libraries do not have a method for getting all currently connected accounts like the browser wallet API has, which is why this was changed. We can consider expanding the libraries to support this if we feel that we need it.

## Checklist

      hard-to-understand areas.


euroe-demo/src/components/BeerStore.tsx
```diff
@@ -71,24 +86,33 @@ async function submitTransaction(items: bigint, account: AccountAddress.Type, wa
   const transferMetadata: CIS2.CreateTransactionMetadata = {
     energy: Energy.create(dryRun.usedEnergy.value + 100n),
   };
-  const transferTx = await client.createTransfer(transferMetadata, transfer);
+  const transferTx = client.createTransfer(transferMetadata, transfer);
+  const typedParams: TypedSmartContractParameters = {
+    parameters: transferTx.parameter.json,
+    // Add missing padding to the base64 encoding. Ideally the method would allow for base64 without
+    // padding, or the CIS2 client would use padding.
```
abizjak, line 93: Could you make an issue for the client to embed the schema that is padded.

I think there should not be a danger of parsing incorrectly if the padding is added so the change should be safe.

Also, since you're updating dapp libraries, could you please also update this
```
function schemaAsBuffer(schemaBase64: string) {
    const res = toBuffer(schemaBase64, 'base64');
    // Check round-trip. This requires the provided schema to be properly padded.
    if (res.toString('base64') !== schemaBase64) {
        throw new Error(`provided schema '${schemaBase64}' is not valid base64`);
    }
    return res;
}
```
to not require padding, something like

```
function schemaAsBuffer(schemaBase64: string) {
    const res = toBuffer(schemaBase64, 'base64');
    // Check round-trip. This requires the provided schema to be properly padded.
   const serialized = res.toString('base64');
    if (serialized.startsWith(schemaBase64)) {
        const padding = serialized.slice(schemaBase64.length);
        if (padding.substring('==')) {
            return res;
        }
    }
    throw new Error(`provided schema '${schemaBase64}' is not valid base64`);
}

abizjak (approved):

euroe-demo/src/components/BeerStore.tsx
orhoj: I have created an issue, and I agree that it should not be a problem to make that change. Any decoder should handle that without issue.

Issue #75, opened 2024-04-04
Track-and-trace dapp improvements


DOBEN:
**Task description**
Summary of improvements for the `track-and-trace` dapp:

**Sub-tasks**
- [x] Display version at the front end.
- [x] Add link at front-end to the source code repo.
- [x] Generalize the state machine so that we can easily add/remove new addresses to be producer/seller/... via the front end.
- [x] Enable updating of the account when switching an account in the wallets
- [ ] Improve error messages shown in the admin pages on the front end.
       - For example when a non-admin tries to add itself as admin:
![image](https://github.com/Concordium/concordium-dapp-examples/assets/5469652/0b937dad-7dda-4fbb-9485-2cc8fecbcbcc)
- [ ] Make navbar not transparent to avoid overlap in text
![image](https://github.com/Concordium/concordium-dapp-examples/assets/13039169/65d30ad5-f729-44c0-8137-791feb677671)
- [ ] Create a migration script that allows us to read out the current state-machine and admins as set in the smart contract and initialize a new smart contract instance with the same state. This enables us to easily roll out new versions of the hosted dapp without manually adding admins to a newly deployed contract.

PR #76, opened 2024-04-04
Add removing and adding of transition edges

DOBEN:
## Purpose

Addresses https://github.com/Concordium/concordium-dapp-examples/issues/75

To simplify the use of this demo. We want to enable productTeam to easily add themself as a privileged role into the state machine.

## Changes
Smart contract changes:
- Add functions to update (add and remove) state transition edges in the smart contract.

Front end changes:
- Add front end page for transition edge management by admin.
- Add link to hosted front end in Read.me file.
- Display front-end version and link to source code in the nav bar.
- Add /Fix `eslint` errors/warnings


trackAndTrace/frontend/src/components/Explorer.tsx
```diff
@@ -103,11 +103,11 @@ export function Explorer() {
             throw Error(`'itemID' input field is undefined`);
         }
 
-        // Clear table.
-        const tableBody = document.getElementById('table');
-        if (tableBody) {
-            tableBody.innerHTML = '';
-        }
+        // // Clear table.
```
abizjak, line 106: Presumably this should be removed?

trackAndTrace/frontend/vite.config.ts
```diff
@@ -84,6 +84,15 @@ const viteConfig: UserConfig = {
...
         },
     },
+    define: {
+        global: 'globalThis',
+        'process.env': {
+            TRACK_AND_TRACE_NETWORK: process.env.TRACK_AND_TRACE_NETWORK,
+            TRACK_AND_TRACE_SPONSORED_BACKEND_API: process.env.TRACK_AND_TRACE_SPONSORED_BACKEND_API,
+            TRACK_AND_TRACE_NODE: process.env.TRACK_AND_TRACE_NODE,
+            TRACK_AND_TRACE_CONTRACT_ADDRESS: process.env.TRACK_AND_TRACE_CONTRACT_ADDRESS,
+        },
```
abizjak, line 94: Why is this necessary?

trackAndTrace/smart-contract/src/lib.rs
```diff
@@ -247,6 +238,8 @@ pub enum CustomContractError {
     WrongEntryPoint, // -14
     /// Failed signature verification: Signature is expired.
     Expired, // -15
+    /// Update of state machine was unsuccessful.
+    UnSuccessful, // -16
```
abizjak, line 242:
```suggestion
    Unsuccessful, // -16
```

trackAndTrace/smart-contract/src/lib.rs
```diff
@@ -326,6 +319,24 @@ impl<S: HasStateApi> State<S> {
         targets.insert(to)
     }
 
+    /// Remove a transition. Return if the transition was removed successfully.
```
abizjak, line 322:
```suggestion
    /// Remove a transition. Return if the transition was present.
```

trackAndTrace/smart-contract/src/lib.rs
```diff
@@ -326,6 +319,24 @@ impl<S: HasStateApi> State<S> {
...
+        from: Status,
+        address: AccountAddress,
+        to: Status,
+    ) -> bool {
+        let mut transition = self
+            .transitions
+            .entry(from)
+            .or_insert_with(|| StatusTransitions {
+                transitions: builder.new_map(),
+            });
```
abizjak, line 335: This seems needless. If the entry is not there should you not just return false immediately?


```rust
        let Some(mut transition) = self
            .transitions
            .get_mut(from) else { return false; };
```
???

trackAndTrace/smart-contract/src/lib.rs
```diff
@@ -636,6 +647,84 @@ fn change_item_status(
...
+        }
+        Update::Remove => {
+            let success = state.remove(
+                state_builder,
+                params.from_status,
+                params.address,
+                params.to_status,
+            );
+
+            ensure!(success, CustomContractError::UnSuccessful);
```
abizjak, line 721: In principle we don't need to do this check. If this returns false it just means there was nothing to remove. So nothing was changed.

trackAndTrace/frontend/vite.config.ts
DOBEN: It is necessary to run the `front-end` locally without the rest of the components. If you follow the read.me, you get an error trying to run `yarn dev` without these definitions.

trackAndTrace/smart-contract/src/lib.rs
DOBEN: I felt it would be better to do it that way, since `productTeam` can now figure out the current state of the `stateMachine` by trying to add/remove an edge via the front end (and observe if the front-end returns an error or not).

Bargsteen (approved):
LGTM. Works as expected :tada:

PR #89, opened 2024-07-17
Update sponsoredTransaction backend with contract client

DOBEN:
## Purpose

related https://github.com/Concordium/concordium-dapp-examples/pull/87

## Changes

- Change to use the `contract-client` from the Rust SDK.
- Add reject reason decoding of reverted transactions during the dry-run.
- `Internal errors` occurring when querying the node are not reported to the front end anymore (might leak some information of the backend node connection) but full logging is done at the backend of any `internal error`.

### DIFF
--- a/sponsoredTransactions/backend/src/handlers.rs
+++ b/sponsoredTransactions/backend/src/handlers.rs
@@ -135,15 +135,16 @@ pub async fn submit_transaction(
 
     log::debug!("Simulate transaction to check its validity.");
 
-    let dry_run = match contract_client
+    let dry_run_result = contract_client
         .dry_run_update_with_reject_reason_info::<PermitParam, LogError>(
             "permit",
             Amount::zero(),
             key.address,
             &param,
         )
-        .await?
-    {
+        .await?;
+
+    let dry_run = match dry_run_result {
         InvokeContractOutcome::Success(dry_run) => Ok(dry_run),
         InvokeContractOutcome::Failure(rejected_transaction) => {
             match rejected_transaction.decoded_reason {
@@ -223,10 +224,7 @@ pub async fn handle_rejection(err: Rejection) -> Result<impl warp::Reply, Infall
         Ok(mk_reply(message.into(), code))
     } else if let Some(LogError::TransactionSimulationError(e)) = err.find() {
         let code = StatusCode::INTERNAL_SERVER_ERROR;
-        let message = format!(
-            "Simulation of transaction rejected with reject reason: {:?}",
-            e
-        );
+        let message = format!("Simulation of transaction rejected with reject reason: {e:?}");
         Ok(mk_reply(message, code))
     } else if let Some(LogError::TransactionSimulationDecodedError(decoded_reason, reject_reason)) =
         err.find()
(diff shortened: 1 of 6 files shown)


sponsoredTransactions/backend/src/handlers.rs
```diff
@@ -145,91 +131,56 @@ pub async fn submit_transaction(
...
-            used_energy: _,
-        } => {
-            log::error!("TransactionSimulationError {:#?}.", info);
-
-            return Err(warp::reject::custom(LogError::TransactionSimulationError(
-                RevertReason {
-                    reason: reason.clone(),
-                },
-            )));
+    let dry_run = match contract_client
```
Victor-N-Suadicani, line 138: suggestion: bind the dry_run_update... call to a variable to make this match a bit more readable

sponsoredTransactions/backend/src/handlers.rs
```diff
@@ -244,63 +195,56 @@ pub async fn submit_transaction(
...
-        let code = StatusCode::INTERNAL_SERVER_ERROR;
-        let message = "Simulation invoke error.";
+        let message = "Cannot access the node.";
         Ok(mk_reply(message.into(), code))
     } else if let Some(LogError::TransactionSimulationError(e)) = err.find() {
         let code = StatusCode::INTERNAL_SERVER_ERROR;
-        let message = format!("Transaction simulation error. Your transaction would revert with the given input parameters: {}", e);
+        let message = format!(
+            "Simulation of transaction rejected with reject reason: {:?}",
+            e
```
Victor-N-Suadicani, line 228:
```suggestion
            "Simulation of transaction rejected with reject reason: {e:?}"
```

sponsoredTransactions/backend/src/handlers.rs
```diff
@@ -244,63 +195,56 @@ pub async fn submit_transaction(
...
+        );
+        Ok(mk_reply(message, code))
+    } else if let Some(LogError::TransactionSimulationDecodedError(decoded_reason, reject_reason)) =
+        err.find()
+    {
+        let code = StatusCode::BAD_REQUEST;
+        let message = format!(
+            "Simulation of transaction rejected in smart contract with decoded reject reason: \
+             `{}` derived from: {:?}.",
+            decoded_reason, reject_reason
```
Victor-N-Suadicani, line 238:
```suggestion
            "Simulation of transaction rejected in smart contract with decoded reject reason: \
             `{decoded_reason}` derived from: {reject_reason:?}."
```

sponsoredTransactions/backend/src/main.rs
```diff
@@ -149,6 +156,11 @@ async fn main() -> anyhow::Result<()> {
 
     log::debug!("Get public files to serve.");
 
+    // Check if the front end has been built and the public folder exists.
+    if !Path::new(&app.public_folder).exists() {
```
Victor-N-Suadicani, line 160: This kinda sounds like something that should be a compile-time check? But not sure

Victor-N-Suadicani (approved):

sponsoredTransactions/backend/src/main.rs
DOBEN:
```
cargo run -- --node http://node.testnet.concordium.com:20000 --port 8080 --account ./4SizPU2ipqQQza9Xa6fUkQBCDjyd1vTNUNDGbBeiRGpaJQc6qX.export --public-folder ../frontend/dist --smart-contract-index 9586
```
The check is about the `--public-folder` flag when you run the backend. So the check is during run-time and not during compile-time.

Victor-N-Suadicani: Ah makes sense.

Issue #96, opened 2024-08-06
Front end for `compliant-reward-distribution-dapp`


DOBEN:
https://www.figma.com/design/2q2Rdmt5FuR0Pi0p6aRPDX/Mobile-Wallet-Master?node-id=1952-4651&t=ABvE3OShW7EVeb36-0

# Subtasks
- [x] Flows for generating the signatures/zk proofs with the Android CryptoX mobile wallet as well as the browser wallet to submit the tweet/zkproof to the backend.
- [x] Flows for admins to read data from the database.
- [x] Add dockerBuild/dockerCompose files to host the front-end. 
- [x] Release pipelines. 
- [  ] Apply Figma UI design.

PR #100, opened 2024-09-02
Initial template

DOBEN:
## Purpose

Addresses https://github.com/Concordium/concordium-dapp-examples/issues/96

Next PR:
- Apply Figma UI design.

## Changes

Initial `compliance-reward-distribution-dapp` front-end.


compliant-reward-distribution/indexer-and-server/src/bin/server.rs
```diff
@@ -95,6 +112,29 @@ struct Args {
...
+    } else {
+        "mainnet"
+    };
+
+    let config = serde_json::json!({
+        "node": node_endpoint.uri().to_string(),
+        "network": network,
+    });
+    let config_string = serde_json::to_string(&config).expect("JSON serialization always succeeds");
+    serde_json::json!({ "config": config_string })
```
Victor-N-Suadicani, line 137: This feels a bit weird, you're constructing a JSON object with a "config" key that is itself a JSON object but in serialized string form? Why not construct it like this
```
serde_json::json!({
    "config": {
        "node": node_endpoint.uri().to_string(),
        "network": network,
    }
});
```
so that it is just a regular JSON object?

compliant-reward-distribution/indexer-and-server/src/bin/server.rs
```diff
@@ -181,6 +235,8 @@ async fn main() -> anyhow::Result<()> {
         .route("/api/canClaim", post(can_claim))
         .route("/api/getZKProofStatements", get(get_zk_proof_statements))
         .route("/health", get(health))
+        .nest_service("/assets", serve_dir_service)
+        .fallback(get(|| async { Html(index_html) }))
```
Victor-N-Suadicani, line 239: Doesn't this fallback make it so you can never get a 404? I.e. /api/nonexistent_method returns the frontend HTML? I guess it's not a big deal but this could be surprising behavior.

Victor-N-Suadicani (approved):
Only checked the Rust part.

compliant-reward-distribution/indexer-and-server/src/bin/server.rs
DOBEN: The `config_string` is injected as a handlebar to the frontend. I have experimented but couldn't find a way to properly inject it as an JSON object.

https://github.com/Concordium/concordium-dapp-examples/pull/100/files#diff-308f9df34b9c6315d0c3664ac1aab85bdff22a88a48c8801b79badf9aeceb7b0R8

Here is the mimic that the frontend does when running it isolated from the backend.
https://github.com/Concordium/concordium-dapp-examples/pull/100/files#diff-5442a8d500c255de0d6a4a35479557f4544c1ae96f758ecc80b71108f40f37a3R85

Victor-N-Suadicani: Ah I see! It would be great to include a comment here to explain why it is done like this, since the code otherwise looks quite strange.

compliant-reward-distribution/indexer-and-server/src/bin/server.rs
```diff
@@ -235,8 +236,43 @@ async fn main() -> anyhow::Result<()> {
...
+            }),
+        )
+        .route(
+            "/",
+            get({
+                let index_html = index_html.clone();
+                move || async { Html(index_html) }
+            }),
+        )
+        .nest_service("/assets", serve_dir_service.clone())
```
soerenbf, line 275: Hmm.. I don't agree with this change 😝 

Is it not possible to make a "sub-router" for any `/api` route (by using [`Router::nest`](https://docs.rs/axum/latest/axum/struct.Router.html#method.nest)) and then still have the fallback? It feels weird that one would have to add routes in 2 places to be able to handle the initial request to any route defined in the frontend application.

Victor-N-Suadicani: I would say usually you wouldn't even have the backend serve the frontend like this. You would just put nginx in front (which you'd do anyway for SSL/HTTPS) and you'd have Nginx route all /api paths to the backend and any other paths to a dedicated frontend server. Then the backend doesn't need to serve the frontend directly like this. But I don't know how this is supposed to be deployed.

DOBEN: @soerenbf The reason the fallback route was removed was that any URL route would otherwise point to the home page from the front end. I don't mind the exact behavior of having a fallback or not but I don't want to change forth and back, so I would like to keep it now.

@Victor-N-Suadicani From a pure dapp perspective a framework such as `next.js` where frontend/backend is written in one language and well integrated would be ideal. Nonetheless, for our dApps we prefer to have a Rust backend (our Rust SDK has been most up-to-date concerning cryptographic functions and support for various new additions to the blockchain, and we had more manpower/expertise in that language over the years so there is more code written that can be re-used). 
One reason why we started to serve the front-end via the backend is that often they have to be set up with matching environmental variables (e.g. use the same smart contract instances, the same network and/or the same node connection (at least a node to the same network).

Victor-N-Suadicani: Not sure if you're misunderstanding or I am :sweat_smile: - I am not saying the frontend and backend should be the same language. I'm just saying a common way to deploy a backend API alongside a server that serves a frontend web application is to use nginx to route API paths (like all paths prefixed with `/api`) to the backend and all other routes serve the frontend application. It just moves the routing from the backend into nginx and makes it so the backend does not need to be aware of the frontend's distribution files and such. This has no bearing on how the frontend/backend are organized in code to share some configuration.

Anyway, it's not important and this probably works fine :)

soerenbf: But we still need to build the config into the frontend application, which is the whole point of this setup IMO. Another nice thing about doing it this way is that we can pass on the **runtime** configuration to the frontend instead of having to build/deploy a new frontend application if parameters change

@DOBEN alright, I understand if that's annoying. It was more of a convenience of not having to define things twice; not a deal-breaker 😄

compliant-reward-distribution/frontend/src/utils.ts
```diff
@@ -0,0 +1,99 @@
...
+    const recentBlockHash = (
+        await grpcClient.client.getBlocksAtHeight({
+            blocksAtHeight: {
+                oneofKind: 'absolute',
+                absolute: {
+                    height: { value: recentBlockHeight },
+                },
+            },
+        })
+    )?.response.blocks[0].value;
```
soerenbf, line 43: why do you use the inner `client` to do this instead of the top level interface of `grpcClient`?

compliant-reward-distribution/frontend/src/wallet-connection.tsx
soerenbf, line 1: So out of curiosity.. why not use our `@concordium/wallet-connectors` or `@concordium/react-components` library for this?

I don't blame you though, I don't think they're very nice to use myself (especially the `react-components package`...) 😅

compliant-reward-distribution/indexer-and-server/src/bin/server.rs
```diff
@@ -597,10 +697,41 @@ async fn post_zk_proof(
...
+    // Create an `uniqueness_hash` to identify the identity associated with the
+    // account by hashing the concatenated string of `national_id` and
+    // `nationality`. Every identity should only be allowed to receive
+    // rewards once (with one of their accounts). The `nationality` is a
+    // two-letter country code (ISO 3166-1 alpha-2).
+    // Note: Concatenating a fixed-size string (`nationality`) with a non-fixed-size
+    // string (`national_id`) is safe. Two non-fixed-size strings would be unsafe.
+    // E.g. `format!("{}{}", "AA", "BB")` and `format!("{}{}", "A", "ABB")`
+    // would produce the same hash even if the strings are different.
+    let concatenated = format!("{}{}", national_id, nationality);
```
soerenbf, line 709: So it's a win for multi-national users. But I guess there's no way around that 🤷 

Do we know if these attributes are always available? Or I guess that would just be a requirement to be able to claim the reward?

compliant-reward-distribution/frontend/src/components/ConnectWallet.tsx
```diff
@@ -0,0 +1,49 @@
...
+                    Android CryptoX Wallet
+                </Button>
+                <br />
+                <Button
+                    variant="primary"
+                    disabled={true}
+                    onClick={async () => connectProvider(await WalletConnectProvider.getInstance())}
+                >
+                    iOS CryptoX Wallet [Coming Soon]
+                </Button>
```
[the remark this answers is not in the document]
DOBEN: The i0S wallets cannot yet produce ZK proof. But yes, you are right that someone with an i0S still can use the above `walletConnect` from the Android button and connect (not sure how we should prevent that, since I am not aware that the walletConnect session reveals the connected device type). So the above UI was chosen to signal to i0S users that they cannot use this dapp for now.

compliant-reward-distribution/frontend/src/wallet-connection.tsx
DOBEN: I think, I got frustrated about these `double-connect` buttons that you have to click to connect your wallet as we have this in some of the dapps. Having this `wallet-connection.tsx` avoided the hassle with the buttons :).
https://wccd.testnet.concordium.com/

soerenbf: Yes.. I ended up doing sort of the same for voting because of the exact same frustration. It would be very nice with some proper design of especially `@concordium/react-components` and perhaps if needed also `@concordium/wallet-connectors`. I feel like they're both overly complex to use.

compliant-reward-distribution/frontend/src/apiReqeuests.ts
```diff
@@ -0,0 +1,290 @@
+import { AccountAddress, AtomicStatementV2, CredentialStatement, VerifiablePresentation } from '@concordium/web-sdk';
+
+interface AccountData {
+    // The account address that was indexed.
+    accountAddress: AccountAddress.Type;
```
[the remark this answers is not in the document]
DOBEN: yes, the json string returned by the backend in a base58 hex string and seems to be parsed into the `AccountAddress.Type` here.

compliant-reward-distribution/frontend/src/components/TweetSubmission.tsx
```diff
@@ -0,0 +1,102 @@
...
+        control: control,
+        name: ['tweet'],
+    });
+
+    async function onSubmit() {
+        setError(undefined);
+        setSuccessfulSubmission(undefined);
+
+        try {
+            checkTweetUrlFormat(tweet);
```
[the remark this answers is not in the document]
DOBEN: Good point.

compliant-reward-distribution/frontend/src/components/ZkProofSubmission.tsx
```diff
@@ -0,0 +1,120 @@
...
+                    <br />
+                    <Button
+                        variant="info"
+                        id="accountAddress"
+                        disabled={true}
+                        hidden={successfulSubmission === undefined}
+                    >
+                        Success
+                    </Button>
+                </Form>
```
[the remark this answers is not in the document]
DOBEN: Replaced the `disabled` Button with a `Alert info` box (makes more sense as you said). 

In terms of using the `Form`, it gives a nice and consistent UI on all pages and allows to validate of the input fields efficiently. It is a bit of an overkill since some of the submissions don't have an input field (in that case a simple Button would be equally possible), but again it is mainly kept for consistency on each page.

compliant-reward-distribution/frontend/src/apiReqeuests.ts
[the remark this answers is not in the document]
soerenbf: Yes.. but how? 😁 it wouldn't just happen automatically?

compliant-reward-distribution/frontend/src/utils.ts
DOBEN: I was wondering why this parameter is so complex without support in our web-SDK :). Will fix it.

compliant-reward-distribution/indexer-and-server/src/bin/server.rs
DOBEN: If people have several IDs (with different data), there is no way around preventing them from claiming for each identity (nonetheless the amount of IDs a person has access to is often rather small). This is a philosophical problem and the solution would be a global/worldwide identity issuing authority (which doesn't exist for now) that issues unique identifiers to everyone.

Discussions about good `context-strings` have been done in this thread and "national_id" + "nationality" was one of their options.
https://support.concordium.software/t/sybil-resilient-identifier-and-proof-of-uniqueness/1359

I think we should ignore edge cases such as identities that don't have "national_id"/"nationality" attributes. If they still can claim (lucky them), if not also not that bad.

compliant-reward-distribution/frontend/src/apiReqeuests.ts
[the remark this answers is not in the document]
DOBEN: One problem identified with `zod` is that it only evaluates and parses primitive javascript types. Our backends often return custom types (e.g. `AccountAddress`, `AtomicV2Statememtn` from our web-sdk) which would need to be rebuild with `zod` primitive types to be able to be parsed.

compliant-reward-distribution/frontend/src/App.tsx
```diff
@@ -0,0 +1,88 @@
...
+                <Route
+                    path="/zkProofSubmission"
+                    element={<ZkProofSubmission prover={account} provider={provider} grpcClient={grpcClient} />}
+                />
+                <Route
+                    path="/tweetSubmission"
+                    element={<TweetSubmission signer={account} provider={provider} grpcClient={grpcClient} />}
+                />
+
+                <Route path="/Admin" element={<Admin signer={account} provider={provider} grpcClient={grpcClient} />} />
```
soerenbf, line 83: Would it be possible to hide this route if the connected account is not an admin? Is that possible to look up through the API? Otherwise it would be a nice addition since I think it's a bit noisy as a regular user.

soerenbf (commented):
Here are some things I think makes sense to change. Mostly minor stuff, but an annoying one is the account change/disconnect not propagating to the application. Would also be nice if admin page was hidden for regular users.

Otherwise it looks good 🥇

compliant-reward-distribution/frontend/src/App.tsx
DOBEN: yes, this route will be hidden from the average user after applying the Figma changes in the next PR. The navigation via the navbar at the top will disappear and instead, buttons will lead the user to the next step (while the admin page will not be part of the user flow). Our admins will still be able to get to the admin page by knowing that the can expand the URL with `/admin`.

soerenbf (approved):
Looks great! 💯

PR #140, opened 2025-07-22
fix: update docker build workflow

nb-ccd:
## Purpose

Currently, every docker image is build in response to any pull request. This updates the Github Actions workflow to only trigger builds when non-markdown files are changed in the target directory.

## Changes

Add paths-filter action to generate a list of changed paths as step output: https://github.com/dorny/paths-filter 

Pass that output into the image build steps as a matrix to trigger only the relevant workflow.

Add conditional steps to set the docker file location and working directory depending on which build is triggered, as well as custom variables for track & trace dapp.

## Checklist

      hard-to-understand areas.

I'm not sure if we track changes to CI in any changelog for this repo, please let me know if so.

### DIFF
--- a/.github/workflows/docker-build.yaml
+++ b/.github/workflows/docker-build.yaml
@@ -6,24 +6,82 @@ on:
 
   pull_request:
     branches: main
+    paths:
+      - 'compliant-reward-distribution/**'
+      - 'euroe-demo/**'
+      - 'signMessage/**'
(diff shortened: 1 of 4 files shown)


mt-concordium (approved):
Looks good to me! Although not sure how do we feel about consuming actions from unknown individuals? Maybe creating a fork would be a way to go, or using a commit SHA rather than tag which can forcibly changed?

PR #142, opened 2025-07-24
feat: update rust lint and test job

nb-ccd:
## Purpose

Updates the Rust components to share a workspace, and rust-ci.yaml workflow to run a single set of linting/validation/build jobs against all crates. Part of https://linear.app/concordium/issue/COR-1722/use-cargo-workpaces

## Changes

* Added workspace and implemented necessary changes in CI, Dockerfiles and crates (some crates had to be renamed to have unique names e.g.)
* Special handling of smart contracts in CI and workspace Cargo.toml, they cannot be part of the workspace due to cargo concordium not supporting workspaces
* Removed the Rust SDK as submodule. The SDK dependencies are now via crates.io
* Use `toolchain.toml` to specify rust version
* Added crate trackAndTrace/common to share types between the indexer and the smart contract for that project (before dependency was directly on the smart contract, which was not nice and did also not compile when using a workspace)


.github/workflows/rust-fmt-lint.yaml
allanbrondum, line 1: Instead of checking which crates to build, we should unify the build in a single workspace. It will be much simpler and just as efficient. In a single workspace, dependencies are only build once (and our own code takes minimal time to build compared to dependencies). Also, you just have one job doing one source code checkout etc. I have a PR doing it for the misc repo: https://github.com/Concordium/concordium-misc-tools/pull/248. I just need to adjust the profile build the smart contract and adjust the paths in the release workflows. We should do something similar for this repo.

.github/workflows/rust-fmt-lint.yaml
```diff
@@ -17,23 +28,90 @@ env:
   CARGO_CONCORDIUM_VERSION: "4.2.0"
```
allanbrondum, line 28: We should remove `RUST_VERSION: 1.85` and replace it with a `rust-toolchain.toml` file instead, then it takes effect for all builds including in development environment.

nb-ccd: Makes sense, will do :)

If that's the version we're testing against now, shall I update version recommendations in the readme? Eg we mention 1.70 here: https://github.com/Concordium/concordium-dapp-examples/blob/main/sponsoredTransactionsAuction/backend/README.md?plain=1#L76

Cargo.toml
```diff
@@ -0,0 +1,39 @@
+[workspace]
+# Not sure if needed, seems like this is the default behaviour as of rust 2021?
+# https://doc.rust-lang.org/edition-guide/rust-2021/default-cargo-resolver.html
+resolver = "2"
+
+members = [
+  "compliant-reward-distribution/indexer-and-server",
+  "gallery/verifier",
+  "low-code-nft-marketplace/cis2-market",
```
nb-ccd, line 9: There's no rust build instructions in the readme here: https://github.com/Concordium/concordium-dapp-examples/tree/main/low-code-nft-marketplace/cis2-market

Should we add instructions equivalent to other repos, eg same rustfmt, clippy and build commands?

Cargo.toml
```diff
@@ -0,0 +1,37 @@
+[workspace]
+resolver = "2"
```
nb-ccd, line 2: Not sure if this setting is needed, it seems like this is the default behaviour as of rust 2021?

https://doc.rust-lang.org/edition-guide/rust-2021/default-cargo-resolver.html

nb-ccd: Gotcha, thanks.

Cargo.toml
```diff
@@ -0,0 +1,39 @@
...
+
+members = [
+  "compliant-reward-distribution/indexer-and-server",
+  "gallery/verifier",
+  "low-code-nft-marketplace/cis2-market",
+  "sponsoredTransactionsAuction/backend",
+  "sponsoredTransactions/backend",
+  "trackAndTrace/smart-contract",
+#  "trackAndTrace/test-scripts",
```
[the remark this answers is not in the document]
allanbrondum: We should move `rustfmt.toml` to the top of the repo, else it will not be picked up. That may mean that some crates that not has this formatting config applied now gets it applied and needs to be reformatted. I see it as a benefit of workspaces that it enforces uniform rules like this, since this is what we want anyways (and have in other repos)

compliant-reward-distribution/indexer-and-server/CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 ## Unreleased changes
 
+## 0.1.3
```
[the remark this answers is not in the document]
allanbrondum: Probably not necessary. Also, it is just and example, so versioning semantics does not really have any impact

compliant-reward-distribution/indexer-and-server/Cargo.toml
```diff
@@ -1,5 +1,5 @@
 [package]
-name = "indexer"
+name = "crdindexer"
```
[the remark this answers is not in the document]
allanbrondum: I think your renaming is ok, it makes things work

.github/workflows/rust-fmt-lint.yaml
```diff
@@ -12,138 +12,94 @@ on:
...
       - name: Download and install Cargo Concordium
         run: |
           curl -L https://github.com/Concordium/concordium-smart-contract-tools/releases/download/releases/cargo-concordium/${CARGO_CONCORDIUM_VERSION}/cargo-concordium-linux-amd64 -o /tmp/cargo-concordium
           chmod +x /tmp/cargo-concordium
           sudo mv /tmp/cargo-concordium /usr/bin/cargo-concordium
 
-      - name: Run cargo concordium test
-        run: cargo concordium test --out "${{ matrix.crates }}/concordium-out/module.wasm.v1" -- --manifest-path "${{ matrix.crates }}/Cargo.toml"
+      - name: Run cargo concordium test against low-code-nft-marketplace cis2-market  
+        run: cargo concordium test --out "low-code-nft-marketplace/cis2-market/concordium-out/module.wasm.v1" -- --manifest-path "low-code-nft-marketplace/cis2-market/Cargo.toml"
```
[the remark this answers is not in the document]
allanbrondum: `cargo concordium test` should be part of the matrix build for the two smart contracts that reside in a single crate workspace

trackAndTrace/smart-contract/Cargo.toml
```diff
@@ -16,15 +16,11 @@ serde = ["concordium-contracts-common/derive-serde", "dep:serde"]
...
+concordium-contracts-common = "9.2.0"
 
 [dev-dependencies]
 concordium-smart-contract-testing = { version = "4.2" }
 rand = "0.8"
 
 [lib]
 crate-type = ["cdylib", "rlib"]
-
-[profile.release]
```
limemloh, line 28: Why was this removed? Did you check the impact on the smart contract?

trackAndTrace/smart-contract/Cargo.toml
```diff
@@ -16,15 +16,11 @@ serde = ["concordium-contracts-common/derive-serde", "dep:serde"]
 concordium-std = { version = "10.0", default-features = false }
 concordium-cis2 = { version = "6.1", default-features = false }
 serde = { version = "1", optional = true, features = ["derive"] }
-concordium-contracts-common = "*"
+concordium-contracts-common = "9.2.0"
```
limemloh, line 19: Since this is reexported by `concordium-std` this was probably made to use `*` so we don't have to sync the versions and the resolver would just use whatever `concordium-std` is using.

.github/workflows/rust-ci.yaml
```diff
@@ -0,0 +1,91 @@
+name: "Rust: build and test"
+
+on:
+  push:
+    branches: main
```
squirer, line 5: i feel like we dont need to specify a push to the main branch here, because any PR that is raised is already going to be scanned with the pull request specified below, as long as protection rules do not allow merge of code that does not pass these rust ci checks we are safe

.github/workflows/typescript-ci.yaml
```diff
@@ -1,20 +1,17 @@
-name: Build, Format and Lint Frontends
+name: "TypeScript: build and test"
 
 on:
   push:
     branches: main
 
   pull_request:
```
squirer, line 7: as you are running on every pull request now, you in theory do not need to run it for main above (same as my comment above) but we probably need to decide if it makes sense to do all checks everywhere everytime as a safety measure

compliant-reward-distribution/dockerfiles/indexer.Dockerfile
```diff
@@ -5,16 +5,15 @@ ARG RUST_IMAGE=rust:1.85-bookworm
 # Build indexer
 FROM ${RUST_IMAGE} as build
 WORKDIR /indexer
-COPY ./compliant-reward-distribution/indexer-and-server ./
-COPY ./deps/concordium-rust-sdk /deps/concordium-rust-sdk
-RUN cargo build --release
+COPY ../ ./
```
squirer, line 8: changing it to this will include the front end from the parent folder, do we intend to have the front end copied into the docker env? There is a danger here that anything else added to the parent folder will also be copied into the docker env if someone is not familiar - just noting this incase that part was missed

compliant-reward-distribution/dockerfiles/server.Dockerfile
```diff
@@ -14,9 +14,10 @@ RUN yarn build
 # Build server
 FROM ${RUST_IMAGE} AS server
 WORKDIR /server
-COPY ./compliant-reward-distribution/indexer-and-server ./
-COPY ./deps/concordium-rust-sdk /deps/concordium-rust-sdk
-RUN cargo build --release
+COPY ../ ./
+# It seems as if the `crdindexer` package outputs both and indexer and a server binary...
+# Is the name just a little misleading?
+RUN cargo build --locked -p crdindexer --release
```
squirer, line 20: I have a wider opinion seeing these docker files all exist in a single directory, i don't think its a good idea - they should live inside the app they are intending to build and conceptually they are responsible for the scope of that

gallery/verifier/Cargo.toml
```diff
@@ -7,6 +7,8 @@ license-file = "../../LICENSE"
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
 
 [dependencies]
+concordium-rust-sdk.workspace = true
+concordium_base = "8.0.0"
```
DOBEN, line 11: If we keep the submodule dependency then it is easier to update if we don't have hard coded versions.

```suggestion
concordium-rust-sdk.workspace = true
concordium_base = "*"
```

gallery/verifier/README.md
```diff
@@ -56,7 +56,7 @@ Changes to any of the packages must be such that
 -   `cargo clippy --all` produces no warnings
 -   `rust fmt` makes no changes.
 
-Everything in this repository should build with stable rust at the moment (at least version 1.62 and up), however the fmt tool must be from a nightly release since some of the configuration options are not stable. One way to run the `fmt` tool is
+Everything in this repository should build with stable Rust at the moment (our CI system uses 1.85), however the fmt tool must be from a nightly release since some of the configuration options are not stable. One way to run the `fmt` tool is
```
DOBEN, line 59: I suggest that we remove the explicit mention of version `1.85` as we will update the version in the future and it is hard to remember to update this READ.me accordingly. We could point to the file where the version is defined instead if we want to keep that info.

sponsoredTransactionsAuction/backend/README.md
```diff
@@ -68,12 +68,13 @@ The smart contract code at {index: AUCTION_CONTRACT_INDEX, subindex: 0} can be f
...
 
 This repository's CI automatically checks formatting and common problems in rust.
+
 Changes to any of the packages must be such that
 
 - ```cargo clippy --all``` produces no warnings
 - ```rust fmt``` makes no changes.
 
-Everything in this repository should build with rust version 1.70 however the `fmt` tool must be from a nightly release since some of the configuration options are not stable. One way to run the `fmt` tool is
+Our CI systems use Rust 1.85, however the `fmt` tool must be from a nightly release since some of the configuration options are not stable. One way to run the `fmt` tool is
```
DOBEN, line 77: Suggestion would be to avoid mention the version in the READ.me as it is hard to maintain that.

trackAndTrace/dockerfiles/indexer.Dockerfile
```diff
@@ -4,10 +4,9 @@ ARG RUST_IMAGE=rust:1.85-bookworm
 
 # Build indexer
 FROM ${RUST_IMAGE} as build
-COPY ./trackAndTrace/smart-contract ./smart-contract
 WORKDIR /indexer
-COPY ./trackAndTrace/indexer ./
-RUN cargo build --release
+COPY . .
```
squirer, line 8: Isnt this copying everything inside the current dockerfiles directory? this seems like it might be incorrect.

trackAndTrace/dockerfiles/server.Dockerfile
```diff
@@ -12,10 +12,9 @@ RUN yarn build
 
 # Build server
 FROM ${RUST_IMAGE} AS server
-COPY ./trackAndTrace/smart-contract ./smart-contract
 WORKDIR /server
-COPY ./trackAndTrace/indexer ./
-RUN cargo build --release
+COPY . .
```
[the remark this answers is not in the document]
DOBEN: The docker files are being executed in the root of the repo folder (based on the given context). 

https://github.com/Concordium/concordium-dapp-examples/blob/main/trackAndTrace/docker-compose.yml

But as you mention the current copying of everything in the root seems not ideal as we want to generate a small image (only with necessary code in it) and there is a potential that something (like `.env` files) get leaked into the image if we copy everything into it.

Cargo.toml
```diff
@@ -0,0 +1,31 @@
...
+  "gallery/verifier",
+  "sponsoredTransactionsAuction/backend",
+  "sponsoredTransactions/backend",
+  "trackAndTrace/test-scripts",
+  "trackAndTrace/indexer",
+  "trackAndTrace/sponsored-transaction-service"
+]
+
+[workspace.dependencies]
+concordium-rust-sdk = "7.0.0"
```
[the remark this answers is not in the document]
allanbrondum: I think it is not a good way to display examples, that we use submodules. I think we discussed it at some standup some time ago and agreed to remove submodules from this repo. They way to test dependents during development is via patches in Rust: https://doc.rust-lang.org/cargo/reference/overriding-dependencies.html#the-patch-section

.github/workflows/rust-ci.yaml
allanbrondum: When we introduce tagging releases via the CI, it will likely also be the main branch trigger that creates a tag after running the CI checks on main.

.github/workflows/docker-build.yaml
```diff
@@ -1,11 +1,10 @@
...
 
 on:
   push:
     branches: main
 
   pull_request:
-    branches: main
     paths:
       - 'compliant-reward-distribution/**'
       - 'euroe-demo/**'
```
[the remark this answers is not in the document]
allanbrondum: Below there is a check on which components have actually changes and which images should be build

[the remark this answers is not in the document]
allanbrondum: (Notice it is just a test that building docker images works, the image is thrown away, it is not a release)

compliant-reward-distribution/dockerfiles/indexer.Dockerfile
allanbrondum: When using workspaces, the entire workspace code is in the temporary builder container, see comment below.

trackAndTrace/dockerfiles/server.Dockerfile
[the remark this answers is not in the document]
allanbrondum: (By the way, I think this PR likely improves what is copied, since before the entire rust sdk submodule (with all its transitive submodules) was copied into the builder stage)

gallery/verifier/README.md
allanbrondum: Agree, I'll clean up the docs a bit, try to make it a little better (but maybe not perfect ;))

trackAndTrace/smart-contract/Cargo.toml
allanbrondum, line 1: fix build of this before merge

trackAndTrace/smart-contract/Cargo.toml
allanbrondum: I think it is even resolved independently of what versions other dependencies use of the crate, it will just resolve to the newest major version, unless locked.

compliant-reward-distribution/dockerfiles/indexer.Dockerfile
allanbrondum: No that copy is a copy into the build stage/build container. The copy into the output image is the line
```
COPY --from=build ./indexer/target/release/crdindexer /crdindexer
```
Or am I overlooking something? :)

PR #241, opened 2026-06-10
feat(allowListManager): Support multiple tokens

andynelson:
## Purpose

We can now manage multiple tokens with configurable proofs on tokens that have an allow list. Each token has it's own configuration for both the amount to airdrop, whether to add the user to the allow list, and the kind of proof required when doing so. (supports set inclusion/exclusion four country fields, and range for dob)

## Changes

Frontend and backend changes to support multiple tokens, with backwards compatibility with the single token version

## Checklist

      hard-to-understand areas.

## CLA acceptance

_Remove if not applicable.

By submitting the contribution I accept the terms and conditions of the
Contributor License Agreement v1.0
- link: https://developers.concordium.com/CLAs/Contributor-License-Agreement-v1.0.pdf

- [x] I accept the above linked CLA.

### DIFF
--- a/allowListManagerFaucet/backend/src/modules/token-distribution/token-distribution.service.ts
+++ b/allowListManagerFaucet/backend/src/modules/token-distribution/token-distribution.service.ts
@@ -138,8 +138,9 @@ export class TokenDistributionService implements OnModuleInit {
    */
   async startTokenDistribution(dto: AddToAllowListDto): Promise<string> {
     const processId = uuidv4()
-    const tokenId = dto.tokenId || this.tokenConfigs[0].id
-    const tokenConfig = this.tokenConfigs.find(c => c.id === tokenId) ?? this.tokenConfigs[0]
+    const bail = (msg: string) => throw new Error(msg)
+    const tokenId = dto.tokenId ?? this.tokenConfigs[0].id
+    const tokenConfig = this.tokenConfigs.find(c => c.id === tokenId) ?? bail(`Could not find configured token with ID ${tokenId}.`)
 
     // Eligible only if the account has no token entry at all (never held this token).
     // A null balance means no entry; any string value (including '0') means the account
--- a/allowListManagerFaucet/backend/src/modules/token-distribution/token-distribution.service.ts
+++ b/allowListManagerFaucet/backend/src/modules/token-distribution/token-distribution.service.ts
@@ -1,5 +1,6 @@
-import { Injectable, Logger } from '@nestjs/common'
+import { Injectable, Logger, OnModuleInit } from '@nestjs/common'
 import { ConfigService } from '@nestjs/config'
+import { readFileSync } from 'node:fs'
 import { ConcordiumService } from '../concordium/concordium.service.js'
 import { ProcessTrackingService } from './process-tracking.service.js'
 import { TokenId, TokenAmount, Token, TokenHolder, CborMemo, Cbor } from '@concordium/web-sdk/plt'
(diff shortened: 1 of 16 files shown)


allowListManagerFaucet/backend/src/modules/token-distribution/token-distribution.service.ts
```diff
@@ -12,66 +13,177 @@ import {
...
+    const tokenId = dto.tokenId || this.tokenConfigs[0].id
+    const tokenConfig = this.tokenConfigs.find(c => c.id === tokenId) ?? this.tokenConfigs[0]
+
+    // Eligible only if the account has no token entry at all (never held this token).
+    // A null balance means no entry; any string value (including '0') means the account
+    // has held the token before, even if the balance is now zero.
+    const currentBalance = await this.getTokenBalance(tokenConfig.id, dto.userAccount)
+    if (currentBalance !== null) {
+      throw new Error(`Account ${dto.userAccount} has previously held token ${tokenConfig.id} and is not eligible for distribution`)
+    }
```
soerenbf, line 151: Don't know how important this is to guard against, but this is not effective against parallel requests being made. If we want to guard against that, we should introduce some kind of lock.

allowListManagerFaucet/backend/src/modules/token-distribution/token-distribution.service.ts
soerenbf, line 1: This is not something that is introduced in this PR, but the service does not actually verify accounts, which means that the distribution endpoint can be invoked by non-verified accounts by using the API directly without using the frontend

allowListManagerFaucet/frontend/vite.config.ts
```diff
@@ -1,20 +1,32 @@
...
+      wasm(),
+      topLevelAwait(),
+      {
+        name: 'inject-runtime-config',
+        transformIndexHtml(html) {
+          return html
+            .replace('__BACKEND_URL__', env.BACKEND_URL || 'http://localhost:3001')
+            .replace('__VERIFIER_URL__', env.VERIFIER_URL || 'http://localhost:7017');
+        },
+      },
```
soerenbf, line 26: This replaces the runtime config done in `build-and-serve.sh` with build-time config.. just something to have in mind. In general I feel like runtime config is favourable 🤷

allowListManagerFaucet/backend/src/modules/token-distribution/token-distribution.service.ts
```diff
@@ -12,66 +13,177 @@ import {
...
-   * Starts the token distribution process for a verified user
-   * Executes all operations in a single atomic transaction
+   * Starts the token distribution process for a verified user.
+   * Resolves the token ID from the DTO (or first configured token) and its configured mint amount.
    */
   async startTokenDistribution(dto: AddToAllowListDto): Promise<string> {
     const processId = uuidv4()
-    const tokenId = dto.tokenId || this.defaultTokenId
+    const tokenId = dto.tokenId || this.tokenConfigs[0].id
+    const tokenConfig = this.tokenConfigs.find(c => c.id === tokenId) ?? this.tokenConfigs[0]
```
soerenbf: Falling back to the first configured token when `tokenId == undefined` is intended behaviour, but silently falling back to the first configured token when a token config could not be found seems like it should error instead.


```suggestion
    const bail = (msg: string) => throw new Error(msg)
    const tokenId = dto.tokenId ?? this.tokenConfigs[0].id
    const tokenConfig = this.tokenConfigs.find(c => c.id === tokenId) ?? bail(`Could not find configured token with ID ${tokenId}.`)
```

soerenbf (approved):

allowListManagerFaucet/backend/src/modules/token-distribution/token-distribution.service.ts
andynelson: Yeah, something we'd probably want to do if this goes beyond devnet use, but I was disinclined to add state without rearchitecting so it has some form of persistence layer

allowListManagerFaucet/backend/src/modules/token-distribution/token-distribution.service.ts
andynelson: true, but that's only an issue if the backend is reachable from the web, rather than only by the frontend

allowListManagerFaucet/frontend/vite.config.ts
andynelson: yeah, I agree - on the list for next time this gets updated

allowListManagerFaucet/backend/src/modules/token-distribution/token-distribution.service.ts
soerenbf: how is it not reachable through any http client if it's reachable by the frontend? 🤔

andynelson: the frontend and backend sit on an internal network, with a load balancer providing access to the frontend from the web

andynelson:
```mermaid
architecture-beta
    service lb(internet)[Load Balancer]
    service fe(server)[Front End] in vpc
    service be(server)[Back End] in vpc
    group vpc(cloud)[Private Network]
   lb: R --> L:fe
   fe: B --> T:be
```

soerenbf: Not sure I follow.. the browser runs the app which calls `/distrubite`, no?

```sh
❯ curl -X POST https://allowlistapi.testnet.concordium.com/token-distribution/distribute \
    -H 'Content-Type: application/json' \
    -d '{
      "userAccount": "4tFVVpFpgiEmSjWjZg5sZQ2oR5yKCpjUgoubZ3sdXXZekaPbm2"
    }'
{"processId":"f5a08642-f3d4-4516-bdb8-3b7666f62db9","status":"processing","steps":[{"step":"Execute Token Distribution","status":"processing","progress":100}]}
```

